PR #1, opened 2015-09-14
Added testing infrastructure and collapsible view.

dsjen:
@andy-armstrong @AlasdairSwan @talbs , this is a first pass at creating our first component for the toolkit--a collapse/expand section.

Does the structure look right?  Currently, components would be in their own directory (e.g. `src/collapsible-view/js/views/collapsible-view.js`).  I have concerns that it's a bit deep, and what I'd imagine is that developers would require the package with something like:

```
require(['edx-ui-tooltip/collapsible'],
        function (CollapsibleView) {
...
```

This would mean that the javascript file would be at the root of the project and a bit muddy.  We could write a gulp task that copies javascript files to a `dist` directory perhaps.

I still need to incorporate configurations with require and bower.  In the meantime, I'm pinning Insights to use a particular commit from this branch so that I can continue pushing forward on some functionality there that uses the collapse/expand functionality.

What do you think about using gulp in conjunction with karma as our test runner?

I'd love to hear your thoughts!

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,0 +1,25 @@
+{
+  "name": "edx-ui-toolkit",
+  "version": "0.1.0",
+  "homepage": "https://github.com/edx/edx-ui-toolkit",
+  "repository": {
+    "type": "git",
+    "url": "git://github.com/edx/edx-ui-toolkit.git"
+  },
+  "ignore": [
+    "**\/.*",
+    "node_modules",
+    "bower_components"
+  ],
+  "license": "Apache 2.0",
+  "dependencies": {
+    "jquery": "~1.11.1",
+    "underscore": "~1.8.2",
+    "backbone": "~1.1.2",
+    "requirejs": "~2.1.15"
+  },
+  "devDependencies": {
+    "requirejs-plugins": "~1.0.3",
+    "jasmine": "~2.0.4"
+  }
+}
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,0 +1,18 @@
+// gulpfile.js
+
+// Rather than manage one giant configuration file responsible
+// for creating multiple tasks, each task has been broken out into
+// its own file in gulp/tasks. Any files in that directory get
+// automatically required below.
+
+// To add a new task, simply add a new task file that directory.
+// gulp/tasks/default.js specifies the default set of tasks to run
+// when you run `gulp`.
+
+(function () {
+    'use strict';
+    var requireDir = require('require-dir');
+
+    // Require all tasks in gulp/tasks, including subfolders
+    requireDir('./gulp/tasks', { recurse: true });
+}());
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -1,0 +1,83 @@
+// Karma configuration
+
+module.exports = function (config) {
+    config.set({
+
+        // base path that will be used to resolve all patterns (eg. files, exclude)
+        basePath: '',
+
+
+        // frameworks to use
+        // available frameworks: https://npmjs.org/browse/keyword/karma-adapter
+        frameworks: ['jasmine', 'requirejs', 'sinon'],
+
+
+        // list of files / patterns to load in the browser
+        files: [
+            {pattern: 'bower_components/**/*.js', included: false},
+            {pattern: 'bower_components/**/*.json', included: false},
+            {pattern: 'components/**/views/*.js', included: false},
+            {pattern: 'test/specs/**/*.js', included: false},
+            {pattern: 'components/config.js', included: true},
+            {pattern: 'test/spec-runner.js', included: true}
+        ],
+
+
+        // preprocess matching files before serving them to the browser
+        // available preprocessors: https://npmjs.org/browse/keyword/karma-preprocessor
+        preprocessors: {
+            'components/**/*.js': ['coverage']
+        },
+
+        // plugins required for running the karma tests
+        plugins:[
+            'karma-jasmine',
+            'karma-requirejs',
+            'karma-phantomjs-launcher',
+            'karma-coverage',
+            'karma-sinon'
+        ],
+
+        // test results reporter to use
+        // possible values: 'dots', 'progress'
+        // available reporters: https://npmjs.org/browse/keyword/karma-reporter
+        reporters: ['progress', 'coverage'],
+
+        coverageReporter: {
+            dir:'build', subdir: 'coverage-js',
+            reporters:[
+                {type: 'html', subdir: 'coverage-js/html'},
+                {type: 'cobertura', file: 'coverage.xml'},
+                {type: 'text-summary'}
+            ]
+        },
+
+        // web server port
+        port: 9876,
+
+
+        // enable / disable colors in the output (reporters and logs)
+        colors: true,
+
+
+        // level of logging
+        // possible values: config.LOG_DISABLE || config.LOG_ERROR || config.LOG_WARN || config.LOG_INFO || config.LOG_DEBUG
+        logLevel: config.LOG_INFO,
+
+
+        // enable / disable watching file and executing tests whenever any file changes
+        autoWatch: true,
+
+
+        // start these browsers
+        // available browser launchers: https://npmjs.org/browse/keyword/karma-launcher
+        // you can also add Chrome or other browsers too
+        browsers: ['PhantomJS'],
+
+        captureTimeout: 60000,
+
+        // Continuous Integration mode
+        // if true, Karma captures browsers, runs the tests and exits
+        singleRun: false
+    });
+};
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,50 @@
+# Python bytecode:
+*.py[co]
+
+# Packaging files:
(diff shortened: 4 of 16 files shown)


andy-armstrong:
FYI @dan-f

andy-armstrong:
Thanks for being the first to jump in, @dsjen! You have lots of good questions. We should probably look at how other toolkits handle their namespaces, and steal what works well for them. I know that Tyler suggested that we look at Semantic UI: http://semantic-ui.com/.

My first thought is that we will need the same paths to work within the library as when included in another project. So, we will want the path to include ``views``, ``models`` etc. I don't know if we want to have those directories at the root, or whether they should be a level down. 

e.g. ``edx-ui-toolkit/views/collapsible-view`` or ``edx-ui-toolkit/collapsible-view/views/collapsible-view``

For single classes the latter seems too verbose, but I think we might want to group more complicated views and models together. Maybe we could start with ``edx-ui-toolkit/components/views/collapsible-view``? That's how we structured the edx-platform components:

https://github.com/edx/edx-platform/tree/master/common/static/common/js/components

It would be good to include tests with this pull request although I realize that means we'll have more to work out before you can merge it.

andy-armstrong:
I think using Karma definitely makes sense for the test runner. I'm happy to go with Gulp too, although I've never used it. It seems that everyone is using it now.

andy-armstrong:
I forget if I sent this around, but this article is a good summary of the different tools that we should be considering:

http://reactkungfu.com/2015/07/the-hitchhikers-guide-to-modern-javascript-tooling/

benpatterson:
@andy-armstrong just noticed this (I think I am still subscribed to this repo), but minor distinction I wanted to mention.... Gulp can be used as the task manager, with karma as the JS test runner. I believe that is how the dashboard and drupal repos are set up. The two can happily co-exist.

andy-armstrong:
@benpatterson I was unclear. I was saying we should definitely use Karma, and then I don't have a strong preference if we use Gulp in addition.

It would be great if you could advise us as to the best way to set up testing of this repo. It sounds like we probably should follow the dashboard, unless you see any challenges with that.

dan-f:
Just throwing in my $.02 - @andy-armstrong I re-read the modern JavaScript tooling page you sent around, and both gulp and karma seem like good choices. I think it's worth comparing karma to mocha. From their marketing sites, Mocha seems a fair bit simpler.

dsjen:
@dan-f thanks for the cents!  Would using mocha mean that we not use jasmine as our unit testing framework?  I wasn't picturing a departure from jasmine (not that it's out of the question).  Was there a mocha test runner that could run other unit testing frameworks?

@andy-armstrong, using the `edx-ui-toolkit/components/views/collapsible-view` directory structure sounds good to me.  I think it's a bit hefty to import the bower package, but we could figure out some way of creating distributions/builds in the future.

I'll start on restructuring the project and adding gulp in.

dan-f:
@dsjen it looks like I mistook mocha for a karma-level test runner based on the description in the [blog post](http://reactkungfu.com/2015/07/the-hitchhikers-guide-to-modern-javascript-tooling/) which was sent around. I don't see a need to depart from jasmine, though mocha does look pretty neat.

dsjen:
No problem!  I was wondering if I missed something.  :)

dsjen:
I've restructured the component directory based on feedback and added in gulp with a single task to run a linter.

I'll be adding in tests next.

I'm documenting commands to execute on the readme, but it's going to definitely need some fleshing out at some point.

benpatterson:
Yeah @andy-armstrong @dan-f mocha basically is a different stack, if you will. It would mean using a different tool for coverage, for example, and, IIRC, it would mean a departure from Jasmine as well.

For testing changes, yes I'd model after dashboard. (If I remember right, the drupal repo is also pretty analogous so I'm being a bit arbitrary here.) I wouldn't be able to contribute much in setting it up right now but am happy to lend a hand with PRs if those come in...

dsjen:
@benpatterson -- thanks!

I've set up the test workflow using `gulp test`, modeled it after the dashboard, and added a first test.  Please let me know what you think.  You'll find npm, bower, and gulp.

bower.json
```diff
@@ -0,0 +1,18 @@
...
+    "node_modules",
+    "bower_components"
+  ],
+  "dependencies": {
+    "requirejs-plugins": "~1.0.3",
+    "jquery": "~1.11.1",
+    "underscore": "~1.8.2",
+    "backbone": "~1.1.2",
+    "requirejs": "~2.1.15",
+    "jasmine": "~2.0.4"
```
talbs: NIT: This may be an ignorant designer speaking, but is this item more of a devDependency or something you'd use in production too?

gulpfile.js
```diff
@@ -0,0 +1,46 @@
+(function () {
```
talbs: More of a larger question outside of this PR... Should we break up this gulpfile to mirror what [we've done in the UXPL](https://github.com/edx/ux-pattern-library/tree/master/gulp) (and @AlasdairSwan and I tried to do with https://github.com/edx/edx-platform/pull/7175)? There has been some maintenance and abstraction benefits for us in maintaining the doc site and all of our Sass, images, templating tasks.

bower.json
```diff
@@ -0,0 +1,18 @@
+{
```
talbs: NIT: worth syncing up some of the detail levels we've done at https://github.com/edx/ux-pattern-library/blob/master/bower.json? Namely, the license and versioning bits.

talbs:
@dsjen, thanks for the good work. This is great to see get off the ground!

I've verified locally that I can install and follow the basic gulp tasks you've set up (everything seems good). Without some UI to test this against, my terminal output looks beautiful:

```
edx-ui-toolkit :: (dsjen/collapsible) » gulp                                                             
[20:53:13] Using gulpfile ~/edX/edx-ui-toolkit/gulpfile.js
[20:53:13] Starting 'test'...
[20:53:13] Starting 'lint'...
INFO [karma]: Karma v0.12.37 server started at http://localhost:9876/
INFO [launcher]: Starting browser PhantomJS
[20:53:14] Finished 'lint' after 365 ms
INFO [PhantomJS 1.9.8 (Mac OS X 0.0.0)]: Connected on socket QHdnP85zho5any1DpCRZ with id 39701607
PhantomJS 1.9.8 (Mac OS X 0.0.0): Executed 1 of 1 SUCCESS (0.005 secs / 0.008 secs)

=============================== Coverage summary ===============================
Statements   : 95.65% ( 44/46 )
Branches     : 100% ( 4/4 )
Functions    : 88.89% ( 8/9 )
Lines        : 95.65% ( 44/46 )
================================================================================
[20:53:17] Finished 'test' after 3.49 s
[20:53:17] Starting 'default'...
[20:53:17] Finished 'default' after 12 μs
```

I left you a few larger questions as well as some NITs. Let me know if you have thoughts about the larger stuff. Assuming you'll take care of the NITs you think are worthwhile, this is good to go from my perspective.

:+1:

bower.json
dsjen: Good catch!  Yes, jasmine should be in devDependencies.

gulpfile.js
dsjen: That's very interesting, and I really like that it'll be more modular and maintainable..  Let me try to break it up into lint, test, and watch tasks.

dsjen:
Hi All!  I've pushed a few updates including:
- updating bower settings
- breaking out gulp tasks into separate task files

As follow up work to make this the edx-ui-toolkit truly read, I suggest some follow up work:
- add instructions on README (at least more than what's minimally there)
- add metadata to bower.json
- add automated testing and quality checks

bower.json
dsjen: Thanks!  I added the license and version (0.1.0) in addition to an actual license and authors files.

talbs:
Thanks for the update, @dsjen. This is looking great from a general gulp/workfow perspective.

lookin-good-gettin-presents

:+1:

components/collapsible-view/views/collapsible-view.js
```diff
@@ -0,0 +1,58 @@
+define(['backbone', 'underscore'],
```
andy-armstrong: Nit: I was thinking of just ``components/views/collapsible-view.js``. I don't like lots of single file directories.

karma.conf.js
```diff
@@ -0,0 +1,83 @@
...
+        // you can also add Chrome or other browsers too
+        browsers: ['PhantomJS'],
+
+        captureTimeout: 60000,
+
+        // Continuous Integration mode
+        // if true, Karma captures browsers, runs the tests and exits
+        singleRun: false
+    });
+};
```
andy-armstrong: Nit: missing newline

andy-armstrong:
Quick question: have you thought about where other static assets would go, e.g. images?

andy-armstrong:
:+1: Looks fantastic, and it is so great to have a starting point!

andy-armstrong:
How do we get GitHub to run the tests, generate coverage etc?

PR #2, opened 2015-09-17
Introduce pagination collection

andy-armstrong:
https://openedx.atlassian.net/projects/UITK/issues/UITK-1

This is my first attempt at bringing the edx-platform's pagination components into the UI Toolkit.

FYI @dsjen @AlasdairSwan @talbs


andy-armstrong:
@dsjen I'm having trouble debugging my unit tests as this is my first time using Karma. Maybe you can help me work out what I'm doing wrong. Thanks!

andy-armstrong:
@dan-f FYI here's my first attempt to get the edx-platform pagination collection into the UI Toolkit. I finally have the tests running but there are a lot of failures. I'm thinking these are mostly due to the upgrade from Backbone.Paginator 1.x to 2.x. If you have some time it would be great to look at this with you to see if it is just issues with the tests or if the collection itself needs some work.

.jshintrc
```diff
@@ -59,7 +59,7 @@
     "shadow"        : false,       // Suppresses warnings about variable shadowing i.e. declaring a variable that had been already declared somewhere in the outer scope.
     "sub"           : false,       // Suppresses warnings about using [] notation when it can be expressed in dot notation.
     "supernew"      : false,       // Suppresses warnings about "weird" constructions like new function () { ... } and new Object;.
-    "validthis"     : true,       // Suppresses warnings about possible strict violations when the code is running in strict mode and you use this in a non-constructor function.
+    "validthis"     : true,        // Suppresses warnings about possible strict violations when the code is running in strict mode and you use this in a non-constructor function.
```
dan-f: Assuming this was accidental

.jshintrc
```diff
@@ -49,7 +49,7 @@
     "lastsemic"     : false,       // Suppresses warnings about missing semicolons, but only when the semicolon is omitted for the last statement in a one-line block.
     "laxbreak"      : false,       // Suppresses most of the warnings about possibly unsafe line breaks in your code.
     "laxcomma"      : false,       // Suppresses warnings about comma-first coding style.
-    "loopfunc"      : false,       // Suppresses warnings about functions inside of loops.
+    "loopfunc"      : true,        // Suppresses warnings about functions inside of loops.
```
dan-f: It looks like [functions inside of loops can lead to some pretty nasty bugs](http://jshint.com/docs/options/#loopfunc). Is it worth it to flip this to `false` rather than correcting each occurrence?

bower.json
```diff
@@ -18,10 +18,13 @@
     "jquery": "~1.11.1",
     "underscore": "~1.8.2",
     "backbone": "~1.1.2",
-    "requirejs": "~2.1.15"
+    "backbone.paginator": "~2.0",
```
dan-f: Woohoo!

components/pagination/collections/paging_collection.js
```diff
@@ -0,0 +1,258 @@
...
+ *
+ * Subclasses may want to override the following properties:
+ *      - url (string): The base url for the API endpoint.
+ *      - isZeroIndexed (boolean): If true, API calls will use page numbers starting at zero. Defaults to false.
+ *      - perPage (number): Count of elements to fetch for each page.
+ *      - queryParams (object): Query parameters for the API call. Subclasses may add entries as necessary. By default,
+ *          a 'sort_order' field is included to specify the field to sort on. This field may be removed for subclasses
+ *          that do not support sort ordering, or support it in a non-standard way. By default filterField and
+ *          sortDirection do not affect the API calls. It is up to subclasses to add this information to the appropriate
+ *          query string parameters in server_api.
```
dan-f: This doc will need to be updated.

components/pagination/collections/paging_collection.js
```diff
@@ -0,0 +1,258 @@
...
+    define(['backbone.paginator'], function (PageableCollection) {
+        var PagingCollection = PageableCollection.extend({
+            initialize: function () {
+                // These must be initialized in the constructor because otherwise all PagingCollections would point
+                // to the same object references for sortableFields and filterableFields.
+                this.sortableFields = {};
+                this.filterableFields = {};
+            },
+
+            isZeroIndexed: false,
```
dan-f: Why default to `false`? It actually looks like the `PageableCollection`'s `state.firstPage` value is 1 (1-based index) by default.

components/pagination/collections/paging_collection.js
```diff
@@ -0,0 +1,258 @@
...
+
+            state: {
+                firstPage: function () { return this.isZeroIndexed ? 0 : 1; },
+                // Specifies the initial page during collection initialization
+                currentPage: function () { return this.isZeroIndexed ? 0 : 1; },
+                perPage: function () { return this.perPage; }
+            },
+
+            queryParams: {
+                page: function () { return this.currentPage; },
```
dan-f: So I think that in `backbone.paginator` 0.8, the `queryParams` attribute mapped server API query params as keys to their request values as the values of the hash. In 2.0, I think this mapping is somewhat flipped. The keys of this hash should be names of keys in the `state` hash, and their values should be the corresponding query string param names.

From the 2.0 README:
```javascript
  // You can configure the mapping from a `Backbone.PageableCollection#state`
  // key to the query string parameters accepted by your server API.
```

components/pagination/collections/paging_collection.js
```diff
@@ -0,0 +1,258 @@
...
+                return this.filterableFields[this.filterField].displayName;
+            },
+
+            /**
+             * Sets the field to sort on. Sends a request to the server to fetch the first page of the collection with
+             * the new sort order. If successful, the collection resets to page one with the new data.
+             * @param fieldName name of the field to sort on
+             * @param toggleDirection if true, the sort direction is toggled if the given field was already set
+             */
+            setSortField: function (fieldName, toggleDirection) {
```
dan-f: I wonder if we could simplify a lot of this sort logic by using the [new sorting functionality](https://github.com/backbone-paginator/backbone.paginator/blob/master/README.rst#sorting) provided by our updated version of the paginator.

karma.debug.conf.js
```diff
@@ -0,0 +1,16 @@
...
+
+module.exports = function (config) {
+    'use strict';
+
+    var commonConfig = require('./karma.common.conf');
+
+    commonConfig(config, {
+        singleRun: false,
+        autoWatch: true,
+        browsers: ['Chrome'],
```
dan-f: Is there a way that we can run automated tests with Firefox? Maybe we could create a karma.ci.conf.js which runs all browsers, not just WebKit-based ones in the CI environment.

dan-f:
@andy-armstrong it's a little tough for me to review this code, since I reviewed and wrote much of it initially :P. I'd be happy to pair debug the test failures though, if you want a second set of eyes!

andy-armstrong:
@AlasdairSwan FYI, this is the pagination library that we just discussed.

andy-armstrong:
@dan-f Thanks for the great comments. I would love to try to pair program this with you if you had some time. @AlasdairSwan is looking to support pagination in his new feature, and it would be great to base it upon this code.

andy-armstrong:
@peter-fogg It looks like DedX will need this feature, so is there any chance you could take it over and help us to land it? I'm swamped with ORA work so don't have any immediate time, but I can advise. @AlasdairSwan has more details if you need them.

peter-fogg:
@andy-armstrong I'll try to take a look at it when I've got time. Is there a specific time this needs to land by?

.jshintrc
peter-fogg: I think flipping this to `false` is in fact the right idea -- we have a few bugs in `spec_helpers.js` that would have been caught if it were! I'll fix those bugs here, and make a PR against edx-platform as well.

peter-fogg: Update: that PR is https://github.com/edx/edx-platform/pull/10465.

.jshintrc
peter-fogg: It keeps all the comments lined up nicely, so I'm keeping it.

components/pagination/collections/paging_collection.js
peter-fogg: Exactly, so we default to one-based as well.

components/pagination/collections/paging_collection.js
peter-fogg: The documentation certainly does seem to say that, and yet changing `queryParams` to what it suggests indicates that the documentation is patently false. For example, if I set
```javascript
queryParams: {
    perPage: 'page_size'
}
```
I see test failures due to the paginator sending a query string of `/url-goes-here?perPage='page_size'`.

karma.debug.conf.js
peter-fogg: It's actually pretty easy to run with Firefox, using [karma-runner/karma-firefox-launcher](https://github.com/karma-runner/karma-firefox-launcher). I'll see about running with Firefox in Travis (and maybe nightly, Aurora, etc.).

components/pagination/collections/paging_collection.js
peter-fogg: Made some small updates; much of this is still OK.

karma.debug.conf.js
peter-fogg: And indeed, af48d85e74b0a0475e30e11536b3bacfc1dbd238 adds config to run on multiple browsers in Travis.

components/pagination/collections/paging_collection.js
peter-fogg: To be honest, I'm a bit wary of changing the `PageableCollection`'s API too much in this work -- assuming the plan is to move this back into edx-platform at some point, I'd rather not make gratuitous modifications which will require changing client code.

peter-fogg:
@dan-f @andy-armstrong I think this is ready; can you give a round of review when you get a chance?

.jshintrc
dan-f: :clap:

.jshintrc
dan-f: Oh I see. Was comparing it to the red negative diff line instead of the one above :P

components/pagination/collections/paging_collection.js
dan-f: Nice, thanks.

components/pagination/collections/paging_collection.js
dan-f: Huh. I must have read this inversely.

components/pagination/collections/paging_collection.js
dan-f: ugh

facepalm

components/pagination/collections/paging_collection.js
dan-f: Agreed, but we shouldn't have to change the `PageableCollection`'s API in order to do so - I was thinking that within the collection object we could call the library's sorting methods.

gulp/tasks/test_ci.js
```diff
@@ -0,0 +1,14 @@
+(function () {
+    'use strict';
+
+    var gulp = require('gulp'),
+        karma = require('karma').server,
+        path = require('path'),
+        configFile = 'karma.ci.conf.js';
+
+    gulp.task('test_ci', function (callback) {
```
andy-armstrong: Is the idea that Travis is going to run more comprehensive tests than the developer will run locally? I wonder if that's a bad idea because then the developer will have a false sense of confidence when the tests pass. @AlasdairSwan what's your feeling about this?

andy-armstrong:
:+1: Thanks for taking this over, @peter-fogg. My plan was to add the pagination component views to this PR (hence the title) but maybe it makes sense to just merge this as is and have a follow-on PR to add the views.

gulp/tasks/test_ci.js
AlasdairSwan: I agree @andy-armstrong, I think Travis should run the same tests as local unless there is a good reason for them to differ that I'm unaware of. It would make debugging more difficult too.

components/pagination/collections/paging_collection.js
dan-f: I now remember the intent of my original comment - I was thinking that we can just re-use that `state.firstPage` rather than add the redundant variable here.

gulp/tasks/test_ci.js
dan-f: This runs the same set of tests, but in multiple browsers.

gulp/tasks/test_ci.js
```diff
@@ -0,0 +1,14 @@
+(function () {
```
dan-f: @peter-fogg this looks like the same file as `gulp/tasks/test.js` with the exception of `'test_ci'` vs. `'test'` on line 9. Is there a way that we can implement local vs CI testing as the same `'test'` task, but parameterize that task's set of browsers, or more generally, its config file?

components/pagination/collections/paging_collection.js
dan-f: Still wondering about this thread.

dan-f:
@peter-fogg this is looking really good! I'm mostly happy with it, but have a couple open questions regarding possible refactors.

components/pagination/collections/paging_collection.js
```diff
@@ -0,0 +1,258 @@
...
+                page_size: function () { return this.perPage; },
+                text_search: function () { return this.searchString ? this.searchString : ''; },
+                sort_order: function () { return this.sortField; }
+            },
+
+            parse: function (response) {
+                this.totalCount = response.count;
+                this.state.currentPage = response.current_page;
+                this.totalPages = response.num_pages;
+                this.start = response.start;
```
dan-f: @andy-armstrong @peter-fogg - do either of you remember what purpose `this.start` served?

andy-armstrong: I think it was the index of the first item in the page. I think it would be safe to remove it at this point, as the only fields that the client should depend on are:

https://openedx.atlassian.net/wiki/display/AC/edX+REST+API+Conventions#edXRESTAPIConventions-7.Pagination

dan-f: That's what I was hoping... I didn't see it referenced anywhere else in
this PR.

peter-fogg: It looks like the `start` field is also unused in the edx-platform version of this code, so I'll remove it.

gulp/tasks/test_ci.js
peter-fogg: Yes; I've added a `--ci` option to enable testing with multiple browsers. It's enabled in `.travis.yml`.

gulp/tasks/test_ci.js
peter-fogg: I like the idea of running tests on multiple browsers, but it seems unnecessary to take the time to test all of them locally on every run. Most bugs will probably be caught by a single run of the test suite, but having extra defense in the form of cross-browser tests seems nice to have.

components/pagination/collections/paging_collection.js
peter-fogg: But `state.firstPage` is defined in terms of `isZeroIndexed`, as is `state.currentPage`. So we need some way to keep those two functions in sync, and `isZeroIndexed` seems to be a reasonable candidate.

components/pagination/collections/paging_collection.js
peter-fogg: The main issue I see with changing the sorting internals is that our sorting API is intentionally different from that of the paginator's. Specifically, our API has all mutating operations set the `isStale` flag, and relies on the client to call `refresh` at the appropriate time (lazy loading). The paginator, though, prefers to eagerly evaluate operations like sorting. Trying to bridge that gap would likely require either more complicated code in our collection, or a change to the API.

components/pagination/collections/paging_collection.js
dan-f: I see. Sounds good!

components/pagination/collections/paging_collection.js
dan-f: I checked, and we never actually make use of this lazy refresh functionality - I couldn't spot anything in the teams code that sets the search/sort but does not immediately refresh. What's the use case?

dan-f:
@andy-armstrong @peter-fogg @AlasdairSwan - I am beginning work on learner analytics which will make use of this component, so I'm going to take over the merging of this PR. As far as I can tell, the only open question is surrounding the laziness of the API's searching/sorting.

dan-f:
@peter-fogg @andy-armstrong - after taking a look at the new `backbone.paginator` API and our Learner Analytics use cases, I pushed up quite a few changes, which I think make this component easier and more consistent to use. Wherever possible I removed code which duplicated `backbone.paginator` functionality. I also made multiple filters work. This necessarily changes the API of this component, but I think it's well worth it. Please review!

components/pagination/collections/paging_collection.js
```diff
@@ -0,0 +1,369 @@
...
+                    },
+                    function () {
+                        self.state.currentPage = oldPage;
+                        deferred.fail();
+                    }
+                );
+                return deferred.promise();
+            },
+
+
```
andy-armstrong: Nit: extra blank line

test/utils/ajax_helpers.js
```diff
@@ -0,0 +1,172 @@
...
+     * to all requests with the specified statusCode.
+     *
+     * TODO (pfogg): removing this for now. Jasmine 2.0 removes the
+     * `after` hook (as well as the ability to get a reference to the
+     * current spec). As a result this function is broken, and it's
+     * not used anywhere regardless. Fixing this is likely as simple
+     * as converting it to work as a decorator/HOF like `fakeRequests`
+     * below.
+     */
+    // fakeServer = function (that, response) {
```
andy-armstrong: Let's just remove this. We can add it back in the future if we really want it.

test/utils/ajax_helpers.js
```diff
@@ -0,0 +1,172 @@
...
+                args = Array.prototype.slice.call(arguments);
+            xhr.onCreate = function(request) {
+                requests.push(request);
+            };
+            spec.apply(null, args.concat([requests]));
+            xhr.restore();
+        };
+    };
+
+    expectRequest = function(requests, method, url, body, requestIndex) {
```
andy-armstrong: Can you add a JavaScript doc string.

test/utils/ajax_helpers.js
```diff
@@ -0,0 +1,172 @@
...
+        if (_.isUndefined(requestIndex)) {
+            requestIndex = requests.length - 1;
+        }
+        request = requests[requestIndex];
+        expect(request.url).toEqual(url);
+        expect(request.method).toEqual(method);
+        expect(request.requestBody).toEqual(body);
+    };
+
+    expectJsonRequest = function(requests, method, url, jsonRequest, requestIndex) {
```
andy-armstrong: Can you add a JavaScript doc string.

test/utils/ajax_helpers.js
```diff
@@ -0,0 +1,172 @@
...
+        request = requests[requestIndex];
+        expect(new URI(request.url).path()).toEqual(expectedUrl);
+        parameters = new URI(request.url).query(true);
+        delete parameters._;  // Ignore the cache-busting argument
+        expect(parameters).toEqual(expectedParameters);
+    };
+
+    /**
+     * Intended for use with POST requests using application/x-www-form-urlencoded.
+     */
```
andy-armstrong: Add parameter doc.

test/utils/ajax_helpers.js
```diff
@@ -0,0 +1,172 @@
...
+        if (_.isUndefined(requestIndex)) {
+            requestIndex = requests.length - 1;
+        }
+        request = requests[requestIndex];
+        expect(request.url).toEqual(url);
+        expect(request.method).toEqual('POST');
+        expect(_.difference(request.requestBody.split('&'), body.split('&'))).toEqual([]);
+    };
+
+    respondWithJson = function(requests, jsonResponse, requestIndex) {
```
andy-armstrong: Another missing doc string.

test/utils/ajax_helpers.js
```diff
@@ -0,0 +1,172 @@
...
+        if (_.isUndefined(requestIndex)) {
+            requestIndex = requests.length - 1;
+        }
+        requests[requestIndex].respond(200,
+            { 'Content-Type': 'application/json' },
+            JSON.stringify(jsonResponse));
+    };
+
+    respondWithError = function(requests, statusCode, jsonResponse, requestIndex) {
```
andy-armstrong: Another missing doc string.

test/utils/ajax_helpers.js
```diff
@@ -0,0 +1,172 @@
...
+        if (_.isUndefined(jsonResponse)) {
+            jsonResponse = {};
+        }
+        requests[requestIndex].respond(statusCode,
+            { 'Content-Type': 'application/json' },
+            JSON.stringify(jsonResponse)
+        );
+    };
+
+    respondWithTextError = function(requests, statusCode, textResponse, requestIndex) {
```
andy-armstrong: Another missing doc string.

test/utils/ajax_helpers.js
```diff
@@ -0,0 +1,172 @@
...
+        if (_.isUndefined(textResponse)) {
+            textResponse = '';
+        }
+        requests[requestIndex].respond(statusCode,
+            { 'Content-Type': 'text/plain' },
+            textResponse
+        );
+    };
+
+    respondWithNoContent = function(requests, requestIndex) {
```
andy-armstrong: Another missing doc string.

test/utils/spec_helpers.js
```diff
@@ -0,0 +1,53 @@
...
+define([], function () {
+    'use strict';
+
+    /**
+     * Runs func as a test case multiple times, using entries from data as arguments. Like Python's DDT.
+     * @param data An object mapping test names to arrays of function parameters. The name is passed to it() as the name
+     *      of the test case, and the list of arguments is applied as arguments to func.
+     * @param func The function that actually expresses the logic of the test.
+     */
+    /* jshint loopfunc:true */
```
andy-armstrong: I'm not sure but I think this might mess up the documentation generation (when we get to it). Can we move it into the body of the function?

andy-armstrong:
Looks great, thanks @dan-f. :+1: (please update doc strings first if possible)

dan-f:
Thanks for the review, @andy-armstrong! I added doc strings to those spec/ajax helper files, which were taken from [edx-platform](https://github.com/edx/edx-platform/blob/master/common/static/common/js/spec_helpers/ajax_helpers.js).

bower.json
```diff
@@ -16,13 +16,16 @@
...
+    "urijs": "1.16.1"
   },
   "devDependencies": {
     "requirejs-plugins": "~1.0.3",
-    "jasmine": "~2.0.4"
+    "jasmine": "~2.0.4",
+    "sinon": "~1.16.1"
   }
-}
+}
```
peter-fogg: Is this auto generated or manually written? If the latter, the file should end in a newline.

components/pagination/collections/paging_collection.js
```diff
@@ -0,0 +1,369 @@
...
+            },
+
+            /**
+             * Returns whether this collection has defined a given
+             * filterable field.
+             * @param fieldName querystring parameter name for the
+             *     filterable field
+             */
+            hasFilterField: function (fieldName) {
+                return !_.isUndefined(this.filterFields[fieldName]);
```
peter-fogg: Though, if you do want the existing behavior this would be cleaner with `_.has`.

components/pagination/collections/paging_collection.js
```diff
@@ -0,0 +1,369 @@
...
+                // metadata (page, size, etc.), and the array of
+                // objects.
+                var modifiedResponse = [];
+                modifiedResponse.push(_.omit(response, 'results'));
+                modifiedResponse.push(response.results);
+                return PageableCollection.prototype.parse.call(this, modifiedResponse, options);
+            },
+
+            /* jshint unused:false */
+            parseState: function (response, queryParams, state, options) {
```
peter-fogg: Is this actually unused? If so, just remove it; otherwise could you leave a comment indicating why it's needed?

peter-fogg: Ah, I see it's inherited from the paginator. I'd add a comment for people not familiar with its API.

bower.json
dan-f: Not sure, but thanks for pointing out. This is one of my biggest pet peeves.

components/pagination/collections/paging_collection.js
dan-f: Good call!

peter-fogg:
:+1: :ship: :shipit: :fireworks:

andy-armstrong:
:+1: 

FYI, I've broken up the JIRA ticket into two since we won't be landing the views. The ticket for merging the PagingCollection: https://openedx.atlassian.net/browse/UITK-1

PR #5, opened 2015-09-18
Introduce new gulp task:  gulp test_debug

andy-armstrong:
This task opens up Chrome so that tests can be debugged using the standard developer tools.

Just do the following instead of using the regular test task:

    gulp test_debug

@dsjen @benpatterson please review.


andy-armstrong:
FYI, I'm using techniques which I got from this web page (which currently will only load from the Google cache):

http://bahmutov.calepin.co/debugging-karma-unit-tests.html

Here's the cached page:

http://webcache.googleusercontent.com/search?q=cache:6kSJpU4-cmkJ:bahmutov.calepin.co/debugging-karma-unit-tests.html&hl=en&gl=us&strip=1&vwsrc=0

I'm surprised that I couldn't find this as a documented technique anywhere else.

gulp/tasks/test_debug.js
```diff
@@ -0,0 +1,26 @@
...
+        path = require('path'),
+        paths = {
+            spec: [
+                'test/specs/**/*.js'
+            ],
+            karamaConf: 'karma.debug.conf.js'
+        };
+
+    // starts up Karma using Chrome in debug mode
+    function runKarma(configFile, cb) {
```
benpatterson: Might just be ignorance on my part, but what kind of object is `cb`? Should that be clearer?

gulp/tasks/test_debug.js
```diff
@@ -0,0 +1,26 @@
+(function () {
+    'use strict';
+
+    var gulp = require('gulp'),
+        karma = require('karma').server,
+        path = require('path'),
+        paths = {
+            spec: [
+                'test/specs/**/*.js'
```
benpatterson: Is there a way to set this spec in a common place to be used by both the usual headless and the debug modes?

karma.conf.js
```diff
@@ -72,7 +73,7 @@ module.exports = function (config) {
         // start these browsers
         // available browser launchers: https://npmjs.org/browse/keyword/karma-launcher
         // you can also add Chrome or other browsers too
-        browsers: ['PhantomJS'],
+        browsers: ['Chrome'],
```
benpatterson: This particular change confuses me. Doesn't this mean headless mode will still run in Chrome?

gulp/tasks/test_debug.js
andy-armstrong: I just copied the default Karma file, but I hate abbreviated names so I'll see if I can find out what it is.

benpatterson: Ok. Not a must.

karma.conf.js
andy-armstrong: No, because I made the test task explicitly pass 'PhantomJS'. I thought that Chrome would be a better default in general, but maybe that's not right.

gulp/tasks/test_debug.js
andy-armstrong: Good question. I'd have to use a similar trick to the debug config, where both files load a file in a relative directory which sets something into the global namespace. My thinking was that there is very little to this configuration and it shouldn't need to change, but I agree that it would be better if it was shared. Let me see what I can come up with.

karma.conf.js
benpatterson: Ah yes, I see that now. In that case, no preference on that change. In terms of browser speed, it goes 1. PhantomJS, 2. Chrome, 3. Firefox. The slower your browser, the more like you are to not-catch flaky scenarios in local development. That's the reason I was originally asking. But I see your change on PhantomJS above, and that's good by me.

gulp/tasks/test_debug.js
benpatterson: I'm afraid I don't have a suggestion. I'd ask @talbs if there's some gulp magick that can be employed, and if not, then let it be.

talbs: Hey, guys. Yes! There is. We currently use a config.js file in the UXPL (https://github.com/edx/ux-pattern-library/blob/master/gulp/config.js) and its how I've set up my personal FED starter (https://github.com/talbs/mise-en-place/blob/master/gulp/config.js).

With each task-based file, we then only set the config and package vars we need and can also call other tasks to run as dependencies (e.g. minifying css https://github.com/talbs/mise-en-place/blob/master/gulp/tasks/css-minify.js).

- - -

On a related note, I ripped a lot of this off of https://github.com/vigetlabs/gulp-starter, which has since been really improved to be a really robust and pluggable static asset pipeline. I'm trying to parse what Viget's doing there now, but could use some help with the heavier levels of JS/abstraction logic. Looks interesting if anyone wants to dig in.

talbs: Actually, a better example of passing settings per spec in is here (https://github.com/talbs/mise-en-place/blob/master/gulp/tasks/css.js#L16), where I'm passing in settings from the CSS portion of my config (https://github.com/talbs/mise-en-place/blob/master/gulp/config.js#L12)

gulp/tasks/test_debug.js
dsjen: I believe `cb` is "callback" and I totally agree that it's unclear.  It's in the gulp documentation at https://github.com/gulpjs/gulp/blob/master/docs/API.md#accept-a-callback.

dsjen:
Another option might be to have the jasmine tests load in the browser (any browser) and be able to debug from there.  I think it'll be a good amount of work to get all the files loaded properly, and I'm doubtful it's really worth it.

gulp/tasks/test_debug.js
andy-armstrong: @talbs Thanks so much for the great links! I haven't really dug into node before, so this is really helpful. I'm going to adopt this technique in all the configuration files.

andy-armstrong:
@dsjen @AlasdairSwan @benpatterson @talbs I've resurrected this PR in case we want to use this during the hackathon. This introduces a new 'test_debug' gulp task that opens tests up into a browser for debugging. Note that once the tests are running you can open them in any browser of your choice by hitting this URL:

http://localhost:9876/debug.html

I've cleaned up the configuration files using a technique I found here:

https://groups.google.com/forum/#!topic/karma-users/uAf7TuVBmCQ

This technique is used by Angular and makes it very simple to have shared configurations with a few overrides. You can see their usage here:

https://github.com/angular/angular.js/blob/master/karma-jqlite.conf.js#L2

karma.headless.conf.js
```diff
@@ -0,0 +1,18 @@
...
+        singleRun: true,
+        autoWatch: false,
+        browsers: ['PhantomJS'],
+        logLevel: config.LOG_INFO,
+        preprocessors: {
+            'components/**/*.js': ['coverage']
+        },
+        reporters: ['progress', 'coverage']
+    });
+};
```
AlasdairSwan: I think it would be better to move this to the common conf file so we don't need a default config, headless and debug (just default (which would be headless) and debug)

dsjen: That's interesting.  Are you thinking that these would be the default values then and other configs would overwrite if needed?

andy-armstrong:
@AlasdairSwan @dsjen I've finally got around to addressing the feedback on this. Let me know what you think.

AlasdairSwan:
Nice :+1:

dsjen:
LGTM! :tada:

andy-armstrong:
Just for the record, I discovered yesterday that ORA2 has a very similar script for running Karma in debug mode:

https://github.com/edx/edx-ora2/blob/master/scripts/js-debugger.sh

PR #8, opened 2015-10-13
[WIP] Migrate LMS card views to the UI Toolkit

andy-armstrong:
https://openedx.atlassian.net/browse/UITK-11

### DIFF
--- a/components/views/card-view.js
+++ b/components/views/card-view.js
@@ -1,0 +1,116 @@
+/**
+ * A generic card view class.
+ *
+ * Subclasses can override any of the following:
+ * - configuration (string or function): Sets the display style of the card as square or list. Can take values of
+ *      "square_card" or "list_card". Defaults to "square_card".
+ * - action (function): Action to take when the action button is clicked. Defaults to a no-op.
+ * - cardClass (string or function): Class name for this card's DOM element. Defaults to the empty string.
+ * - pennant (string or function): Text of the card's pennant. No pennant is displayed if this value is falsy.
+ * - title (string or function): Title of the card. Defaults to the empty string.
+ * - description (string or function): Description of the card. Defaults to the empty string.
+ * - details (array or function): Array of child views to be rendered as details of this card. The class "meta-detail"
+ *      is automatically added to each rendered child view to ensure appropriate styling. Defaults to an empty list.
+ * - actionClass (string or function): Class name for the action DOM element. Defaults to the empty string.
+ * - actionUrl (string or function): URL to navigate to when the action button is clicked. Defaults to the empty string.
+ * - actionContent: Content of the action button. This may include HTML. Default to the empty string.
+ */
+;(function (define) {
+    'use strict';
+    define(['jquery',
+            'underscore',
+            'backbone',
+            'text!templates/views/card-view/basic-card-view.underscore',
+            'text!templates/views/card-view/square-card-view.underscore'],
+        function ($, _, Backbone, basicCardViewTemplate, squareCardViewTemplate) {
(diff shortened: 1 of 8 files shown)


andy-armstrong:
FYI @peter-fogg @dan-f @dsjen @AlasdairSwan

components/views/card-view.js
```diff
@@ -0,0 +1,116 @@
...
+ */
+;(function (define) {
+    'use strict';
+    define(['jquery',
+            'underscore',
+            'backbone',
+            'text!templates/views/card-view/basic-card-view.underscore',
+            'text!templates/views/card-view/square-card-view.underscore'],
+        function ($, _, Backbone, basicCardViewTemplate, squareCardViewTemplate) {
+            var CardView = Backbone.View.extend({
```
AlasdairSwan: Why assign the extend to a variable and return the variable instead of just returning the extend i.e.
```
return Backbone.View.extend({...
```

components/views/card-view.js
```diff
@@ -0,0 +1,116 @@
...
+
+                action: function () { },
+                cardClass: '',
+                pennant: '',
+                title: '',
+                description: '',
+                details: [],
+                actionClass: '',
+                actionUrl: '',
+                actionContent: ''
```
AlasdairSwan: As a convention I think it is easier to read if the variables are declared before any of the functions (i.e. lines 104-111 being above the constructor). Thoughts @andy-armstrong @dsjen ?

components/views/card-view.js
```diff
@@ -0,0 +1,116 @@
...
+                                 this.switchOnConfiguration('square-card', 'list-card') + ' ' +
+                                 this.callIfFunction(this.cardClass);
+                    if (this.callIfFunction(this.pennant)) {
+                        result += ' has-pennant';
+                    }
+                    return result;
+                },
+
+                render: function () {
+                    var maxLength = 72,
```
AlasdairSwan: It would be more extendable if this was a View variables.

components/views/card-view.js
```diff
@@ -0,0 +1,116 @@
...
+                        pennant: this.callIfFunction(this.pennant),
+                        title: this.callIfFunction(this.title),
+                        description: description,
+                        action_class: this.callIfFunction(this.actionClass),
+                        action_url: this.callIfFunction(this.actionUrl),
+                        action_content: this.callIfFunction(this.actionContent),
+                        configuration: configuration,
+                        srInfo: this.srInfo
+                    }));
+                    var detailsEl = this.$el.find('.card-meta');
```
AlasdairSwan: Variables should be declared at the top of functions (because hoisting)

components/views/card-view.js
```diff
@@ -0,0 +1,116 @@
...
+            'underscore',
+            'backbone',
+            'text!templates/views/card-view/basic-card-view.underscore',
+            'text!templates/views/card-view/square-card-view.underscore'],
+        function ($, _, Backbone, basicCardViewTemplate, squareCardViewTemplate) {
+            var CardView = Backbone.View.extend({
+                tagName: 'li',
+
+                events: {
+                    'click .action' : 'action'
```
AlasdairSwan: Currently this isn't doing anything because the cards are links. If we were to update and add an action other than a link then we should use a button instead of an anchor tag. Is there a current use case for cards doing something different on click other than linking to a page?

AlasdairSwan:
As discussed with @andy-armstrong let's hold off on this one until we have a larger discussion around course / discovery cards across all edX platforms (e.g. marketing site vs teams)

PR #9, opened 2015-10-14
Added user menu view and dependencies

AlasdairSwan:
added jquery-karma and tests for the componenet
@andy-armstrong and @dsjen please review
cc @talbs


components/views/dropdown-menu-view.js
```diff
@@ -0,0 +1,231 @@
...
+
+            analyticsLinkClick: function( event ) {
+                var $link = $(event.target),
+                    label = $link.hasClass('user-title') ? 'Dashboard' : $link.html().trim();
+
+                /**
+                 *  Add your own analytics tracking here
+                 *  for example:
+                 */
+                window.analytics.track( 'user_dropdown.clicked', {
```
dsjen: Was the plan to turn this into an option?

bower.json
```diff
@@ -15,10 +15,11 @@
...
   "license": "Apache 2.0",
   "dependencies": {
-    "jquery": "~1.11.1",
-    "underscore": "~1.8.2",
     "backbone": "~1.1.2",
-    "requirejs": "~2.1.15"
+    "jquery": "~1.11.1",
+    "requirejs": "~2.1.15",
+    "text": "~2.0.14",
+    "underscore": "~1.8.2"
```
dsjen: Nice ordering! :)

components/views/dropdown-menu-view.js
```diff
@@ -0,0 +1,231 @@
...
+                    this.$dropdownButton.focus();
+                } else {
+                    items.eq(next).focus();
+                }
+            },
+
+            openMenu: function( $el ) {
+                var $menu = this.$menu;
+
+                if ( !$menu.hasClass( 'is-visible' ) ) {
```
dsjen: Just a suggestion: I tend to like if/else statements in the positive--e.g.:

```
if ( $menu.hasClass( 'is-visible' ) ) {
  this.closeDropdownMenus();
} else {
  ...
}
```

components/views/dropdown-menu-view.js
```diff
@@ -0,0 +1,231 @@
...
+                }
+            },
+
+            listenForPageKeypress: function() {
+                this.$page.on( 'keydown', _.bind( this.escKeypressHandler, this ) );
+            },
+
+            nextMenuItemLink: function( $el ) {
+                var items = this.$el.find('.dropdown-menu').children('.dropdown-item').find('.action'),
+                    items_count = items.length -1,
```
dsjen: It seems like `items_count` should be in camelCase rather.

components/views/dropdown-menu-view.js
```diff
@@ -0,0 +1,231 @@
+define([
```
dsjen: A short description about this component would be helpful.

dsjen:
I've wrapped up my review.  LGTM--just a question about `window.analytics.track`.  I'd really love to use this component!

components/views/dropdown-menu-view.js
AlasdairSwan: For tests to run I kept this in as a default (otherwise it wouldn't be easy to determine whether analytics tracking was working).

components/views/dropdown-menu-view.js
AlasdairSwan: I was following the edX convention of using _ for JSON keys.

components/views/dropdown-menu-view.js
AlasdairSwan: I like your positivity :)

components/templates/dropdown.underscore
```diff
@@ -0,0 +1,13 @@
+<a href="<%- main.url %>" class="user-title">
+    <span class="sr"><%- main.screenreader_label %> </span><%- main.text %>
+</a>
+<button type="button" class="button button-more has-dropdown js-dropdown-button" aria-haspopup="true" aria-expanded="false" aria-controls="edx-user-menu">
+    <span class="sr"><%- button_label %></span>
+</button>
+<ul class="dropdown-menu is-hidden" id="edx-user-menu" tabindex="-1">
+    <% _.each(items, function(item, index) { %>
+        <li class="dropdown-item<% if (items.length-1 === index) { %> last<% } %>">
```
andy-armstrong: Nit: JSHint would reject ``item.length-1`` so it would be good to be consistent by adding the spaces.

components/views/dropdown-menu-view.js
andy-armstrong: That's a good point. We need to work out which JS doc generation tool we're going to use, and where it would look for class level doc. I see that @AlasdairSwan has an excellent doc comment below (maybe added in response to your comment) which maybe is the better place to have it.

components/views/dropdown-menu-view.js
```diff
@@ -0,0 +1,262 @@
...
+         *              url: 'learner_profile'
+         *          }, {
+         *              text: 'Sign Out',
+         *              url: 'logout'
+         *          }
+         *      ]
+         *  parent: 'selecter for parent element that will be replaced with dropdown menu'
+         *
+         */
+        return Backbone.View.extend({
```
andy-armstrong: Nit: I know you had the opposite comment for me in my PR, but in the edx-platform we bind our views to a variable which we return. The benefit of this is that it is easier to search for the class name, because then it finds the definition as well as the usages. We should agree on which style to use...

components/views/dropdown-menu-view.js
```diff
@@ -0,0 +1,262 @@
...
+
+            events: {
+                'click .js-dropdown-button': 'clickOpenDropdown',
+                'click a': 'analyticsLinkClick',
+                'keydown': 'viewKeypress'
+            },
+
+            menu: '.dropdown-menu',
+
+            key: {
```
andy-armstrong: Can we put these into a common utility class? I'm sure we're going to use these a lot. In edx-platform we use the JQuery UI constants, but obviously we don't want to bring that monstrosity in...

components/views/dropdown-menu-view.js
```diff
@@ -0,0 +1,262 @@
...
+
+            postRender: function() {
+                this.$menu = this.$el.find('.dropdown-menu');
+                this.$page = $(document);
+                this.$dropdownButton = $('.js-dropdown-button');
+                this.listenForPageKeypress();
+            },
+
+            analyticsLinkClick: function( event ) {
+                /**
```
andy-armstrong: I don't understand this comment. Are you suggesting that folks should override this class to add the tracking? They won't be modifying this code directly, right?

components/views/dropdown-menu-view.js
```diff
@@ -0,0 +1,262 @@
...
+
+                $open.removeClass( 'is-visible' )
+                     .addClass( 'is-hidden' );
+
+                this.$dropdownButton
+                    .removeClass( 'is-active' )
+                    .attr( 'aria-expanded', 'false' );
+            },
+
+            escKeypressHandler: function( event ) {
```
andy-armstrong: Nit: I think of key press as two words, and I hate abbreviations, so I'd call this ``escapeKeyPressHandler``.

AlasdairSwan: I think the actual JS event is called 'keypress' (https://developer.mozilla.org/en-US/docs/Web/Events/keypress) and the jQuery function is one word too.

test/specs/components/views/dropdown-menu-view-spec.js
```diff
@@ -0,0 +1,168 @@
...
+            it( 'should exist', function () {
+                expect( view ).toBeDefined();
+            });
+
+            it( 'should open the user menu on click of the button', function () {
+                var $btn = view.$el.find('.js-dropdown-button');
+
+                $btn.click();
+
+                setTimeout(function() {
```
andy-armstrong: Can you explain why you need to use ``setTimeout`` for all these tests? I've never used this in Jasmine, and have always relied on everything being synchronous. I've always found that the click event happens before the next line of code, and the timeout idiom makes these tests much more verbose and harder to read. It would be good for me to know why this is necessary.

andy-armstrong: Is this to do with animations? I would have thought that the class was set immediately and the animation was handled by the CSS.

AlasdairSwan: Some TravisCI race condition can occur so that the class is not added in time for the test to pass. This is to prevent this. I'm also not a fan of setTimeout but it is using the jasmine clock.

andy-armstrong:
:+1: Looks fantastic. Can you give me a link to somewhere I can see it too (CodePen or whatever)?

components/utils/constants.js
```diff
@@ -0,0 +1,16 @@
...
+    return {
+        keyCodes: {
+            tab: 9,
+            enter: 13,
+            esc: 27,
+            up: 38,
+            down: 40
+        }
+    };
+});
```
andy-armstrong: Nit: missing newline

components/views/dropdown-menu-view.js
```diff
@@ -0,0 +1,261 @@
+define([
+        'backbone',
+        'jquery',
+        'underscore',
+        '../utils/constants',
+        'text!../templates/dropdown.underscore'
+    ],
+    function( Backbone, $, _, constants, DropdownTpl ) {
```
andy-armstrong: Nit: I'm thinking about what the variable should be called. In edx-platform we would name this ``Constants`` or ``ToolkitConstants`` or something. I'm fine with going with it this way too for now, but I figured I'd at least raise the thought.

andy-armstrong:
:+1: on the new constants... thanks for doing that!

PR #10, opened 2015-10-14
Added spec-reporter for tests that outputs each describe and test tit…

AlasdairSwan:
…le and whether it passes or fails to the console
@dsjen and @andy-armstrong please review
cc @talbs


dsjen:
:+1:

PR #13, opened 2015-11-23
Add JSCS code style validation

andy-armstrong:
I've added JSCS validation to the UI Toolkit, and fixed up the validation errors. I decided to base our rules on the Google presets, but let me know if you disagree with any of this.

Note: I'm introducing JSCS because JSHint has deprecated all of their code style validations. They are switching to linting for code correctness only. See http://jshint.com/docs/options/#maxlen

@AlasdairSwan @dsjen Please review.

FYI @peter-fogg @dan-f @talbs 

P.S. I added JSCS validation with the same set of rules to ORA2: https://github.com/edx/edx-ora2/pull/794


AlasdairSwan:
I'm not a huge fan of forcing ```someFunction( variable );``` to be ```someFunction(variable);```. But for consistency I may be convinced that it's okay. I just find it easier to read longer, chained lines when there is more space within the ();

dan-f:
:+1: I like the idea of standardizing style, especially in a centralized toolkit repository. It makes new code easier to approach by making it more familiar. I prefer no extra padding inside of parens.

gulp/tasks/jscs.js
```diff
@@ -0,0 +1,21 @@
+(function() {
+    'use strict';
+
+    var jshint = require('gulp-jshint'),
```
benpatterson: hey sorry for the drive-by @andy-armstrong , but as an interested linter, I looked at this PR. Should this file be called jshint.js? or, alternatively, should this statement be more like `var jscs = require('gulp-jscs')` ?

andy-armstrong:
@AlasdairSwan It looks like all the built-in presets have no spaces inside parenthesis enabled. You can see them all here:

https://github.com/jscs-dev/node-jscs/search?utf8=%E2%9C%93&q=disallowSpacesInsideParentheses

My preference is to break chained functions onto separate lines for clarity, although I know what you mean about it getting harder to read.

AlasdairSwan:
@andy-armstrong fair point about keeping individual lines more simple, I can live with it :+1:

gulp/tasks/jscs.js
andy-armstrong: You're always welcome, @benpatterson! Nice catch. I did this too quickly and forgot to clean this up. It was intended to all be named for JSRC.

I've squashed and pushed up a fix.

benpatterson:
@andy-armstrong should we consider this kind of change for edx-platform, then? thanks for being on top of this.

andy-armstrong:
@benpatterson Yes, I would like to see us do this for edx-platform too.

dsjen:
:+1: I like the effort to standardize and these makes sense to me.  I actually like the spacing inside of parentheses and have done that in the past, but don't follow that style these days.

andy-armstrong:
@AlasdairSwan @dsjen Any ideas why this fails on Travis but works locally for me? I'm thinking it is a JSHint version issue but I've failed to force it to the right version somehow.

dsjen:
@andy-armstrong -- that's odd.  It looks like it's failing on the bower install step.  2.6.0 (https://github.com/jscs-dev/node-jscs/releases) was released 5 days ago, so there shouldn't be a latency issue...  I'll restart it just in case.

dsjen:
@andy-armstrong -- I think I see the issue.  https://github.com/markelog/node-jscs/releases is a fork of node-jscs and doesn't have a v2.6.0.

andy-armstrong:
@dsjen I tried to downgrade but that didn't seem to work. Would you be able to try fixing this since you're much more adept with Travis than I am? Thanks!

dsjen:
@andy-armstrong I think your fix worked, but there are now some linting errors.  I think this is good!

andy-armstrong:
@dsjen Ah, I see. The linting errors were what I was trying to solve in the first place before I broke the dependencies! I don't get these errors locally, and they seem wrong to me. It makes me think it is using a different version of JSHint, but I don't know how to fix that.

dsjen:
@andy-armstrong -- could we switch back to the node-jscs rather than that fork?  That way we could upgrade to 2.6.0 instead.

dan-f:
@andy-armstrong @dsjen - see https://github.com/edx/edx-ui-toolkit/pull/14

dan-f:
@andy-armstrong - I merged #14. Go ahead and rebase :)

PR #14, opened 2015-11-25
Fix recent travis failures

dan-f:
This should fix the build errors seen in https://github.com/edx/edx-ui-toolkit/pull/13

Background:
I fixed this same bug - in a different way - in Insights [last week](https://github.com/edx/edx-analytics-dashboard/pull/388). Our travis' version of `node`, and consequentially it's versions of `npm` and `semver`, is out of date. The old version of `semver` is permitting prerelease-taged packages to satisfy version requirements, which is why `jshint` 2.9.1-rc1 is getting installed. That version of `jshint` adds backwards-incompatible changes restricting the way that the `'strict'` and `'globalstrict'` options are used. Hence our builds fail. This fixes both the issue with `node` as well as the one with `jshint`.

@benpatterson @andy-armstrong @dsjen

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -44,7 +44,6 @@
     "expr"          : false,       // Suppresses warnings about the use of expressions where normally you would expect to see assignments or function calls.
     "funcscope"     : false,       // Suppresses warnings about declaring variables inside of control structures while accessing them later from the outside.
     "gcl"           : false,       // Makes JSHint compatible with Google Closure Compiler.
-    "globalstrict"  : false,       // Suppresses warnings about the use of global strict mode.
     "iterator"      : false,       // Suppresses warnings about the __iterator__ property.
     "lastsemic"     : false,       // Suppresses warnings about missing semicolons, but only when the semicolon is omitted for the last statement in a one-line block.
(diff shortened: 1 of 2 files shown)


.jshintrc
```diff
@@ -44,7 +44,6 @@
     "expr"          : false,       // Suppresses warnings about the use of expressions where normally you would expect to see assignments or function calls.
     "funcscope"     : false,       // Suppresses warnings about declaring variables inside of control structures while accessing them later from the outside.
     "gcl"           : false,       // Makes JSHint compatible with Google Closure Compiler.
-    "globalstrict"  : false,       // Suppresses warnings about the use of global strict mode.
```
dan-f: Note to reviewers: this is an out-of-date option which is [getting deprecated](http://jshint.com/docs/options/#globalstrict). By removing the option, we still don't tolerate global strict mode (no semantic change), but protect ourselves against future build failures.

dan-f:
For sake of documentation - this was [the GH issue that helped me track down the bug](https://github.com/jshint/jshint/issues/2765#issuecomment-157332792).

andy-armstrong:
Thanks so much @dan-f! This has been driving me nuts for a couple of days. :+1:

dsjen:
Nice! :+1:

PR #16, opened 2015-11-30
Add travis status to README

dan-f:
@andy-armstrong @dsjen @AlasdairSwan

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 Part of [edX code](http://code.edx.org/)
 
-edX UI Toolkit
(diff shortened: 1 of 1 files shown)


andy-armstrong:
:+1: Nice, thanks @dan-f.

Since you're here, do you know how to add a license to the project? Whenever I build it tells me that it is missing. Do we have a standard license we should be using?

AlasdairSwan:
:+1:

dan-f:
@andy-armstrong - see https://github.com/edx/edx-ui-toolkit/pull/17. We're already using the Apache 2.0 license, but it wasn't explicitly called out in the `package.json`. That PR should make the warning go away.

PR #18, opened 2016-01-12
Correctly merges 'queryParams'/'state' in 'extend'

dan-f:
I was improperly handling the way that `backbone.paginator` expects the `state` and `queryParams` properties to be set when overriding with the `extend` function. This change allows child classes to override those properties (recursively) by simply specifying them in the object passed to `extend`, as expected.

@andy-armstrong @AlasdairSwan please review

### DIFF
--- a/components/pagination/collections/paging_collection.js
+++ b/components/pagination/collections/paging_collection.js
@@ -35,28 +35,24 @@
             searchString: '',
             filterableFields: {},
 
+            state: {
+                firstPage: 1,
+                pageSize: 10,
+                sortKey: null
+            },
+
+            queryParams: {
+                currentPage: 'page',
+                pageSize: 'page_size',
+                sortKey: 'order_by',
+                order: 'sort_order',
+                text_search: function () { return this.searchString || null; }
+            },
+
             constructor: function (models, options) {
-                var defaultOptions = {
-                    state: {
-                        firstPage: 1,
-                        pageSize: 10,
-                        sortKey: null
-                    },
-                    queryParams: {
-                        currentPage: 'page',
-                        pageSize: 'page_size',
-                        sortKey: 'order_by',
-                        order: 'sort_order',
-                        text_search: function () { return this.searchString || null; }
-                    }
-                };
-
-                if (_.isUndefined(options)) { options = {}; }
-                _.extend(defaultOptions, _.omit(options, 'state', 'queryParams'));
-                _.extend(defaultOptions.state, options.state);
-                _.extend(defaultOptions.queryParams, options.queryParams);
-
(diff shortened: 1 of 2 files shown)


components/pagination/collections/paging_collection.js
```diff
@@ -377,6 +373,9 @@
                 }
             }
         });
+
+        var PagingCollectionProto = PagingCollection.prototype;
```
AlasdairSwan: Nitpick: variables should be declared at the top of a function.

test/specs/components/pagination/paging_collection_spec.js
```diff
@@ -46,12 +46,22 @@ define(['jquery',
...
                 server.isZeroIndexed = false;
                 server.count = 43;
             });
 
+            it('correctly merges state and queryParams in the extend statement', function () {
+                var MyPagingCollection = PagingCollection.extend({
+                    state: {pageSize: 25},
+                    queryParams: {pageSize: 'per_page'}
+                });
+                var newCollection = new MyPagingCollection([]);
```
AlasdairSwan: Nitpick: there should only be 1 ```var``` per function, using commas to define multiple variables.

AlasdairSwan:
Just a few nitpicks but looks good :+1:

components/pagination/collections/paging_collection.js
dan-f: I figured that since the `PagingCollectionProto` is the prototype of the `PagingCollection`, it made sense to declare it afterwards.

AlasdairSwan: It's more the 2 ```var``` declarations I mean. If you add a comma to line 375 instead of a ```;``` then you could remove the ```var``` from this line.

andy-armstrong: I definitely find it confusing to define it below the lines that are using it. If only we could use ES2015's ``let``!

peter-fogg: I also strongly support declaring this at the top of the file.

dan-f: I'm all for declaring variables at the top of the function, but this is technically declaring the variable at the "top of the function", since it's the second statement in the function following another variable declaration. I could replace the `var` with the comma, but the indentation just makes it look sort of weird.

dan-f: Alternatively, I'll separate the declaration from the assignment.

peter-fogg: Not to take the nitpick train too much further, but you could also just remove the `PagingCollectionProto` entirely :rocket:

dan-f:
Merging this once tests pass.

PR #19, opened 2016-02-12
Refactor dependencies into npm

dan-f:
Allows us to distribute our package on npm.

Also upgrades all the libraries to their latest versions.

[UITK-64](https://openedx.atlassian.net/browse/UITK-64)

@andy-armstrong @cahrens


bower.json
```diff
@@ -8,24 +8,9 @@
   },
   "ignore": [
     "**\/.*",
-    "bower_components",
     "gulp/tasks",
     "node_modules",
     "test"
   ],
-  "license": "Apache 2.0",
-  "dependencies": {
```
andy-armstrong: Does this affect the building of the Bower package? How does Bower know what to include?

andy-armstrong:
@dan-f This looks great. Is there a way to verify that the Bower package will still build correctly once this merges? If you're comfortable that this will work then :+1:

dan-f:
@andy-armstrong - I'm not too familiar with Bower package distribution, but I'll look into it!

Also - I upgraded all the libraries, including jQuery which is now at 2.2.0! :D

dan-f:
@andy-armstrong I was being silly by removing the `dependencies` key/value from the bower.json. I've put it back, which should ensure that our library installs correctly through bower.

cahrens:
:+1:

PR #27, opened 2016-03-09
Release 0.4.0

dan-f:
* Add `getActiveFilterFields` method to the `PagingCollection`

@andy-armstrong

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 0.4.0 (2016-03-09)
+* Add `getActiveFilterFields` method to the `PagingCollection`
+
+- - -
+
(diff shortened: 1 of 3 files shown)


andy-armstrong:
:+1: 

I was considering that 0.3.1 might be a better version number since this is such a minor change, and because otherwise we're going to get to 0.27.0 pretty fast. However, I guess we'll be moving to 1.0 before too much longer, so I'll go with it.

dan-f:
Yeah I wasn't quite sure which way to go. For what it's worth, the semver site suggests incrementing by minor version until 1.0.0 is hit: http://semver.org/#how-should-i-deal-with-revisions-in-the-0yz-initial-development-phase

PR #30, opened 2016-03-11
Implement HtmlUtils helper functions

andy-armstrong:
https://openedx.atlassian.net/browse/UITK-74

This PR adds a new HtmlUtils class to the UI Toolkit, with two functions:
 - ``HTML(text)`` - wraps a known piece of HTML so that it doesn't get double escaped
 - ``interpolate(string, options)`` - returns an HTML-escaped string where the interpolation escapes each option unless it is an instance of an HTMLSnippet (the result of calling HTML)

Note that we have gone around and around on the name of the helper function. I think it is important that it include the word ``escape`` as its behavior is to return an HTML-escaped string, so I ended up going with Robert's original suggestion.

Once this version of the UI Toolkit has been released, I'll make another PR against edx-platform which will bring in this function and replace some or all usages of ``StringUtils.interpolate_text`` with this new, safer function. We will then update the XSS guidelines to advise folks to use this function instead of any of the alternatives, i.e. interpolate, interpolate_text and _.template.

@AlasdairSwan @robrap please review.

FYI @dan-f @nedbat @cahrens

### DIFF
--- a/src/js/utils/html-utils.js
+++ b/src/js/utils/html-utils.js
@@ -1,0 +1,145 @@
+/**
+ * Useful functions for dealing with HTML.
+ *
+ * In particular, these functions default to being safe against
+ * Cross Site Scripting (XSS) attacks.
+ */
+(function(define) {
+    'use strict';
+    define(['underscore', './string-utils.js'], function(_, StringUtils) {
+        var escape, interpolateHtml, HTML, template;
+
+        /**
+         * Creates an HTML snippet.
+         *
+         * The intention of an HTML snippet is to capture a string that
+         * is known to contain HTML that has been safely escaped.
+         * As an example, this allows interpolate to understand that
+         * it does not need to further escape this HTML.
+         *
+         * @param {string} htmlString The string of HTML.
+         * @constructor
+         */
+        function HtmlSnippet(htmlString) {
+            this.text = htmlString;
+        }
+        HtmlSnippet.prototype.valueOf = function() {
+            return this.text;
+        };
+        HtmlSnippet.prototype.toString = function() {
+            return this.text;
+        };
+
+        /**
+         * Helper function to create an HTML snippet.
+         *
+         * @param {string} htmlString The string of HTML.
+         * @returns {HtmlSnippet} An HTML snippet that can be safely rendered.
+         * @constructor
+         */
+        HTML = function(htmlString) {
+            return new HtmlSnippet(htmlString);
+        };
+
+        /**
+         * Returns HTML that is appropriately escaped.
+         *
+         * If the input is provided as a string, then an HtmlSnippet
+         * is returned with the original text escaped. If the input
+         * is already an HtmlSnippet then it is known to be already
+         * escaped and so it is just returned directly.
+         *
+         * @param {(string|HtmlSnippet)} text Some HTML that is
+         * either provided as a string or as an HtmlSnippet.
+         * @returns {HtmlSnippet} A safely escaped HtmlSnippet.
+         */
+        escape = function(text) {
+            if (text instanceof HtmlSnippet) {
+                return text;
+            } else {
+                return HTML(_.escape(text));
+            }
+        };
+
+        /**
+         * Returns an HTML snippet by interpolating the provided parameters.
+         *
+         * The HTML text is provided as a tokenized format string where parameters
+         * are indicated via curly braces, e.g. 'Hello {name}'. These tokens are
+         * replaced by the parameter value of the same name.
+         *
+         * Parameter values will be rendered using their toString methods and then
+         * HTML-escaped. The only exception is that instances of the class HTML
+         * are rendered without escaping as their contract declares that they are
+         * already valid HTML.
+         *
+         * Example:
+         *   HtmlUtils.interpolateHtml(
+         *       'You are enrolling in {spanStart}{courseName}{spanEnd}',
+         *       {
+         *           courseName: 'Rock & Roll 101',
+         *           spanStart: HtmlUtils.HTML('<span class="course-title">'),
+         *           spanEnd: HtmlUtils.HTML('</span>')
+         *       }
+         *   );
+         *
+         * returns:
+         *   'You are enrolling in <span class="course-title">Rock &amp; Roll 101</span>'
+         *
+         * Note: typically the formatString will need to be internationalized, in which
+         * case it will be wrapped with a call to an i18n lookup function. In Django,
+         * this would look like:
+         *
+         *   HtmlUtils.interpolateHtml(
+         *       gettext('You are enrolling in {spanStart}{courseName}{spanEnd}'),
+         *       ...
+         *   );
+         *
+         * @param {string} formatString The string to be interpolated.
+         * @param {Object} parameters An optional set of parameters to the template.
+         * @returns {HtmlSnippet} The resulting safely escaped HTML snippet.
+         */
+        interpolateHtml = function(formatString, parameters) {
+            var result = StringUtils.interpolate(
+                escape(formatString).toString(),
+                _.mapObject(parameters, escape)
+            );
+            return HTML(result);
+        };
+
+        /**
+         * Returns a function that renders an Underscore template as an HTML snippet.
+         *
+         * Note: This helper function makes the following context parameters
+         * available to the template in addition to those passed in:
+         *
+         *   * HtmlUtils: the HtmlUtils helper class
+         *
+         * @param {string} text
+         * @param {object} settings
+         * @returns {function} A function that returns a rendered HTML snippet.
+         */
+        template = function(text, settings) {
+            var HtmlUtils = this,
+                rawTemplate = _.template(text, settings),
+                template = function(data) {
+                    var augmentedData = _.extend(
+                        {
+                            HtmlUtils: HtmlUtils
+                        },
+                        data
+                    );
+                    return HTML(rawTemplate(augmentedData));
+                };
+            return template;
+        };
+
+        return {
+            escape: escape,
+            HTML: HTML,
+            HtmlSnippet: HtmlSnippet,
+            interpolateHtml: interpolateHtml,
+            template: template
+        };
+    });
+}).call(this, define || RequireJS.define);
--- a/src/js/utils/specs/html-utils-spec.js
+++ b/src/js/utils/specs/html-utils-spec.js
@@ -1,0 +1,108 @@
+define(
+    [
+        '../../utils/spec-helpers/spec-helpers.js',
+        '../html-utils.js'
+    ],
+    function(SpecHelpers, HtmlUtils) {
+        'use strict';
+
+        describe('HtmlUtils', function() {
+            describe('HtmlSnippet', function() {
+                it('can convert to a string', function() {
+                    expect(new HtmlUtils.HtmlSnippet('Hello, world').toString()).toBe('Hello, world');
+                });
+            });
+
+            describe('HTML', function() {
+                it('returns an HtmlSnippet', function() {
+                    expect(HtmlUtils.HTML('Hello, world') instanceof HtmlUtils.HtmlSnippet).toBeTruthy();
+                });
+            });
+
+            describe('escape', function() {
+                SpecHelpers.withData({
+                    'HTML escapes text strings': [
+                        'Rock & Roll',
+                        'Rock &amp; Roll'
+                    ],
+                    'HTML escapes full HTML strings': [
+                        '<a href="world">Rock &amp; Roll</a>',
+                        '&lt;a href=&quot;world&quot;&gt;Rock &amp;amp; Roll&lt;/a&gt;'
+                    ],
+                    'does not escape HTML snippets': [
+                        HtmlUtils.HTML('<a href="world">Rock &amp; Roll</a>'),
+                        '<a href="world">Rock &amp; Roll</a>'
+                    ]
+                }, function(input, expectedString) {
+                    var result = HtmlUtils.escape(input);
+                    expect(result.toString()).toBe(expectedString);
+                });
+            });
+
+            describe('interpolateHtml', function() {
+                it('can interpolate a string with no parameters provided', function() {
+                    expect(HtmlUtils.interpolateHtml('Hello, world').toString()).toBe(
+                        'Hello, world'
+                    );
+                });
+
+                SpecHelpers.withData({
+                    'can interpolate a string with empty parameters': [
+                        'Hello, world', {},
+                        'Hello, world'
+                    ],
+                    'can interpolate a string with one parameter': [
+                        'Hello, {name}', {name: 'Andy'},
+                        'Hello, Andy'
+                    ],
+                    'does not interpolate additional curly braces': [
+                        'Hello, {name}. Here is a { followed by a }', {name: 'Andy'},
+                        'Hello, Andy. Here is a { followed by a }'
+                    ],
+                    'escapes characters in the template': [
+                        'Rock & Roll', {},
+                        'Rock &amp; Roll'
+                    ],
+                    'does not escape HTML parameters': [
+                        'Hello, {anchor}', {anchor: HtmlUtils.HTML('<a href="world">world</a>')},
+                        'Hello, <a href="world">world</a>'
+                    ],
+                    'escapes characters in parameters': [
+                        'I love {name}', {name: 'Rock & Roll'},
+                        'I love Rock &amp; Roll'
+                    ],
+                    'does not double escape when chaining interpolate calls': [
+                        'I love {name}', {name: HtmlUtils.interpolateHtml('Rock & Roll')},
+                        'I love Rock &amp; Roll'
+                    ],
+                    'full example': [
+                        'You are enrolling in {spanStart}{courseName}{spanEnd}',
+                        {
+                            courseName: 'Rock & Roll',
+                            spanStart: HtmlUtils.HTML('<span class="course-title">'),
+                            spanEnd: HtmlUtils.HTML('</span>')
+                        },
+                        'You are enrolling in <span class="course-title">Rock &amp; Roll</span>'
+                    ]
+                }, function(template, options, expectedString) {
+                    var result = HtmlUtils.interpolateHtml(template, options);
+                    expect(result.toString()).toBe(expectedString);
+                });
+            });
+
+            describe('template', function() {
+                it('returns a template that renders correctly', function() {
+                    var template = HtmlUtils.template('Hello, <%- name %>'),
+                        result = template({name: 'world'});
+                    expect(result instanceof HtmlUtils.HtmlSnippet).toBeTruthy();
+                    expect(result.toString()).toBe('Hello, world');
+                });
+
+                it('adds HtmlView as an additional context variable for the template', function() {
+                    var template = HtmlUtils.template('I love <%= HtmlUtils.escape("Rock & Roll") %>');
+                    expect(template().toString()).toBe('I love Rock &amp; Roll');
+                });
+            });
+        });
+    }
+);
--- a/src/js/utils/specs/string-utils-spec.js
+++ b/src/js/utils/specs/string-utils-spec.js
@@ -1,0 +1,37 @@
+define(
+    [
+        '../../utils/spec-helpers/spec-helpers.js',
+        '../string-utils.js'
+    ],
+    function(SpecHelpers, StringUtils) {
+        'use strict';
+
+        describe('StringUtils', function() {
+            describe('interpolate', function() {
+                it('can interpolate a string with no parameters provided', function() {
+                    expect(StringUtils.interpolate('Hello, world').toString()).toBe(
+                        'Hello, world'
+                    );
+                });
+
+                SpecHelpers.withData({
+                    'can interpolate a string with empty parameters': [
+                        'Hello, world', {},
+                        'Hello, world'
+                    ],
+                    'can interpolate a string with one parameter': [
+                        'Hello, {name}', {name: 'Andy'},
+                        'Hello, Andy'
+                    ],
+                    'does not interpolate additional curly braces': [
+                        'Hello, {name}. Here is a { followed by a }', {name: 'Andy'},
+                        'Hello, Andy. Here is a { followed by a }'
+                    ]
+                }, function(template, options, expectedString) {
+                    var result = StringUtils.interpolate(template, options);
+                    expect(result.toString()).toBe(expectedString);
+                });
+            });
+        });
+    }
+);
--- a/src/js/utils/string-utils.js
+++ b/src/js/utils/string-utils.js
@@ -1,0 +1,59 @@
+/**
+ * Useful functions for dealing with strings.
+ */
+(function(define) {
+    'use strict';
+    define(['underscore', './string-utils.js'], function(_, StringUtils) {
+        var interpolate;
+
+        /**
+         * Returns a string created by interpolating the provided parameters.
+         *
+         * The HTML text is provided as a tokenized format string where parameters
+         * are indicated via curly braces, e.g. 'Hello {name}'. These tokens are
+         * replaced by the parameter value of the same name.
+         *
+         * Parameter values will be rendered using their toString methods and then
+         * HTML-escaped. The only exception is that instances of the class HTML
+         * are rendered without escaping as their contract declares that they are
+         * already valid HTML.
+         *
+         * Example:
+         *   HtmlUtils.interpolate(
+         *       'You are enrolling in {spanStart}{courseName}{spanEnd}',
+         *       {
+         *           courseName: 'Rock & Roll 101',
+         *           spanStart: HtmlUtils.HTML('<span class="course-title">'),
+         *           spanEnd: HtmlUtils.HTML('</span>')
+         *       }
+         *   );
+         *
+         * returns:
+         *   'You are enrolling in <span class="course-title">Rock &amp; Roll 101</span>'
+         *
+         * Note: typically the formatString will need to be internationalized, in which
+         * case it will be wrapped with a call to an i18n lookup function. In Django,
+         * this would look like:
+         *
+         *   HtmlUtils.interpolate(
+         *       gettext('You are enrolling in {spanStart}{courseName}{spanEnd}'),
+         *       ...
+         *   );
+         *
+         * @param {string} formatString The string to be interpolated.
+         * @param {Object} parameters An optional set of parameters to the template.
+         * @returns {HtmlSnippet} The resulting safely escaped HTML snippet.
+         */
+        interpolate = function(formatString, parameters) {
+            return formatString.replace(/{\w+}/g,
+                function(parameter) {
+                    var parameterName = parameter.slice(1,-1);
+                    return String(parameters[parameterName]);
+                });
+        };
+
+        return {
+            interpolate: interpolate
+        };
+    });
+}).call(this, define || RequireJS.define);
--- a/test/karma.ci.conf.js
+++ b/test/karma.ci.conf.js
@@ -1,6 +1,6 @@
 // Karma configuration to run in Travis
 
-module.exports = function (config) {
+module.exports = function(config) {
     'use strict';
 
     var baseConfig = require('./karma.conf');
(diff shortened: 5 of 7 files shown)


edx-webhook:
## PR Cover Letter

## Sandbox

## Testing
- [ ] i18n
- [ ] RTL
- [ ] Analytics
- [ ] Performance

## Reviewers
- [ ] Code review #1
- [ ] Code review #2
- [ ] UX review
- [ ] Accessibility review
- [ ] Product review

## DevOps assistance
- configuration changes
- feature flags
- migrations
- [ ] File a support ticket

## Post-review
- [ ] Squash commits

src/js/utils/string-utils.js
```diff
@@ -0,0 +1,91 @@
...
+
+    var interpolateEscape, escapeInterpolateOptions, HTML;
+
+    /**
+     * Creates an HTML snippet.
+     *
+     * @param {string} text The HTML for the text.
+     * @constructor
+     */
+    function HTMLSnippet(text) {
```
AlasdairSwan: This function hasn't been defined as a var.

src/js/utils/string-utils.js
```diff
@@ -0,0 +1,91 @@
...
+     *
+     * @param {string} string The string to be interpolated.
+     * @param {Object} options An optional set of options to the template.
+     * @returns {HTML} The resulting HTML-escaped string.
+     */
+    interpolateEscape = function(string, options) {
+        var escapedOptions = escapeInterpolateOptions(options),
+            result = _.escape(string).replace(/{\w+}/g,
+                function(parameter) {
+                    return String(escapedOptions[parameter.slice(1,-1)]);
```
AlasdairSwan: Can you explain what the negative end slice is doing here?

src/js/utils/specs/string-utils-spec.js
```diff
@@ -0,0 +1,58 @@
...
+        '../../utils/spec-helpers/spec-helpers.js',
+        '../string-utils.js'
+    ],
+    function(SpecHelpers, StringUtils) {
+        'use strict';
+
+        describe('StringUtils', function() {
+            describe('interpolateEscape', function() {
+                it('can interpolate a string with no parameters provided', function() {
+                    expect(StringUtils.interpolateEscape('Hello, world').toString()).toBe(
```
robrap: I think this test can be removed, because it duplicates the first withData test.

src/js/utils/string-utils.js
```diff
@@ -0,0 +1,91 @@
...
+    HTMLSnippet.prototype.toString = function() {
+        return this.text;
+    };
+
+    HTML = function(text) {
+        return new HTMLSnippet(text);
+    };
+
+    /**
+     * Returns an HTML-escaped text string by substituting values into
```
robrap: Maybe start with (or include) the following:
Returns an HTMLSnippet. ...

src/js/utils/string-utils.js
robrap: I think these are just dropping the `{` and `}` that are wrapping the parameter name.

AlasdairSwan: So ```-1``` is basically ```length-1```

src/js/utils/string-utils.js
```diff
@@ -0,0 +1,91 @@
...
+     * by the attribute of the same name from the specified options.
+     *
+     * Note: values will be rendered using their toString methods and then
+     * HTML-escaped. The only exception is that instances of the class HTML
+     * are rendered without escaping as their contract declares that they are
+     * already valid HTML.
+     *
+     * Example:
+     *   StringUtils.interpolateEscape(
+     *       'You are enrolling in {spanStart}{courseName}{spanEnd}',
```
robrap: Could you add the gettext() call into this example code.  If you find it necessary, you could add a note that `gettext()` is not strictly needed for interpolateEscape() to work, but it is a requirement for following the best practice of translating strings properly.  I'd just hate to have people copy and paste a code-snippet that doesn't do the right thing.

src/js/utils/string-utils.js
robrap: Yes.  That's not as clear as it should be in [this documentation](http://www.w3schools.com/jsref/jsref_slice_string.asp), but you can see it as a "Tip" and the final example.

robrap:
@andy-armstrong This looks great.  Only some very minor comments.

src/js/utils/string-utils.js
```diff
@@ -0,0 +1,91 @@
...
+     * escaped, but otherwise it will be returned with its value escaped.
+     *
+     * @param {Object} options The options passed to interpolateEscape.
+     * @returns {Object} An object with each option appropriately escaped.
+     */
+    escapeInterpolateOptions = function(options) {
+        var property, value,
+            result = {};
+        for (property in options) {
+            if (hasOwnProperty.call(options, property)) {
```
dan-f: Using `_.mapObject` might make this a bit more readable as it only iterates over an object's own properties, and the iteratee callback gets the key/value pair (no need to index into the object). It also returns a new object which removes the need to manually manage the `result` object.

dan-f:
I know you all are probably sick of hearing me push back against names, but I think they're important, especially for a core API like this. I'm still not on board with `interpolateEscape` because I feel that it highlights an implementation detail over the actual use case. From a developer (user) perspective, I want to interpolate some HTML strings, and the escaping should Just Work™.

From this perspective, I'd like to propose `interpolateHtml` as the new name for this method because that's what we're actually doing; interpolating an HTML string, where escaping - an implementation detail - is handled by the implementation of the API.

andy-armstrong:
@dan-f Naming is both important and hard so I appreciate your perseverance. :-) I initially proposed ``interpolateHtml`` yesterday when I started this, but @robrap convinced me that other JS functions don't have HTML in the name e.g. ``_.escape`` is HTML escaping, but it is focused on the action not the content type. 

One aspect to this is that ``interpolateHtml`` reads like an action, while ``interpolateEscape`` isn't really even a phrase. The former seems a more pleasing name, which must count for something. :-) In Java, the basic escape function is called [escapeHTML](https://commons.apache.org/proper/commons-lang/javadocs/api-2.6/org/apache/commons/lang/StringEscapeUtils.html#escapeHtml(java.lang.String)) which does seem clearer, because you can escape URLs or other formats too.

All this is to say that I think I'm now with you @dan-f . @robrap @AlasdairSwan @nedbat do you have strong preferences either way? Any other name that you think would work better?

src/js/utils/specs/string-utils-spec.js
andy-armstrong: It is subtly different because it tests the case where the options parameter isn't even provided. I tried to think of a way to incorporate this into the other tests but it seemed simpler to keep it separate.

src/js/utils/string-utils.js
andy-armstrong: The way I have this right now, HTMLSnippet is an internal class, so I didn't want to document it. I suppose I could make it public and then the function would be typed more appropriately.

dan-f:
Thanks for the thoughtful response, @andy-armstrong! +1 on `interpolateEscape` not being a phrase.

I see where you're coming from on the whole not calling out the content type, but I also like the idea of eventually having interpolation functions for URLs; we could have a `StringUtils.interpolateURI` which has a similar API but uses a `URI` constructor and wraps up `encodeURIComponent`.

src/js/utils/string-utils.js
andy-armstrong: As we discussed, gettext is a Django i18n function, so it is a little strange to embed that into a general purpose JavaScript library that might be used in IDAs that are not Django based. I'm going to leave the example as is, but then add a description about the needs of i18n.

src/js/utils/string-utils.js
andy-armstrong: That is so much easier! I didn't know about this function before (and amazingly there are no usages of it in any of the platform's code that I can see).

andy-armstrong:
@dan-f After trying this refactoring, I'm now tempted to rename the file to ``html-utils.js`` since all of the functions are to do with managing HTML. It would then look like ``HtmlUtils.interpolateHtml`` and ``HtmlUtils.HTML('<a href="#"></a>')``. Thoughts?

andy-armstrong:
@dan-f pointed out to me that if the class is HtmlUtils then we don't need the suffix in the name. So how about ``HtmlUtils.interpolate``? The name seems unambiguous that this function is the correct interpolation function to use when handling HTML.

dan-f:
I like `HtmlUtils.interpolate` & `HtmlUtils.Html`.

src/js/utils/specs/string-utils-spec.js
robrap: I see.  Then maybe below you can change the comment to "...with empty parameters".

src/js/utils/string-utils.js
robrap: I think HTMLSnippet needs to be external.  

We should agree on its name as well.  As you can see in my comment on the "[Implement in the platform ticket](https://openedx.atlassian.net/browse/UITK-75?focusedCommentId=176420&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-176420)", code outside of this method may want to check its type (String or HTMLSnippet), and determine whether or not it still needs to escape.

robrap:
I'm fine with `HtmlUtils.interpolate` & `HtmlUtils.HTML`.  @dan-f We intentionally decided to go with all-caps for the `HTML` method because we liked how it looked better, but it may be different here if it is following `HtmlUtils`. @nedbat Thoughts?
Also, is everyone fine with `HtmlUtils.HtmlSnippet`, or are there opinions on that as well?

dan-f:
I prefer `Html` over `HTML` since 1) we're distinguishing between JS name, not a general acronym, and 2) it's easier to type. In fact, I now wonder if `HtmlUtils.Html` should actually be `HtmlUtils.html` since it's a factory function rather than a constructor.

I don't mind `HtmlUtils.HtmlSnippet` though I also like `HtmlUtils.HtmlString` as it's a bit shorter and a little bit more specific. I don't mind the distinction here though.

Sorry for being so much trouble! :)

dan-f:
OK I have another idea. Sorry folks. Really! Sorry for the churn!

I want to challenge whether we need both a `HTML` factory function and an `interpolateHtml` function. Instead, I'll propose a polymorphic function named `Html`. With a single argument, it returns that string (just like the `HTML` factory currently does). With a second interpolation argument, it does the interpolation:

```javascript
var Html = require('html-utils').Html;

Html('Please click on {linkStart}this awesome link{linkEnd}.', {
    linkStart: Html('<a href="/the-awesome-page/">'),
    linkEnd: Html('<a href="/the-awesome-page/">')
})

Html('Also check out not-prone we are to XSS: {htmlText}', {
    htmlText: '<script>alert("This will not open an alert.")</script>'
})
```

robrap:
I think we need 2 calls.  It does not make sense to me that this would escape the ``&``:
```
Html('Please click on {linkStart}this great & awesome link{linkEnd}.', {
    linkStart: Html('<a href="/the-awesome-page/">'),
    linkEnd: Html('</a>')
```
But, this would not escape the ``&``:
```
Html('This is a great & awesome example.')
``` 

Also, we went for capitalization because we wanted it to be more declarative.  It just feel more right to me to use `Html('<a>')` than `html('<a>') in this case, even though we are forcing the capitalization a bit.

andy-armstrong:
@robrap @dan-f @AlasdairSwan Thanks for the reviews. I've pushed up a commit which addresses most of the feedback. We will need to chat more on Monday to settle upon the name of the HTML factory method. I've currently left it as it was.

src/js/utils/html-utils.js
```diff
@@ -0,0 +1,115 @@
...
+         * Parameter values will be rendered using their toString methods and then
+         * HTML-escaped. The only exception is that instances of the class HTML
+         * are rendered without escaping as their contract declares that they are
+         * already valid HTML.
+         *
+         * Example:
+         *   HtmlUtils.interpolate(
+         *       'You are enrolling in {spanStart}{courseName}{spanEnd}',
+         *       {
+     *           courseName: 'Rock & Roll 101',
```
dan-f: Looks like these lines got accidentally de-indented.

src/js/utils/html-utils.js
```diff
@@ -0,0 +1,115 @@
...
+                } else {
+                    return _.escape(value);
+                }
+            });
+        };
+
+        return {
+            interpolate: interpolate,
+            HTML: HTML,
+            HtmlSnippet: HtmlSnippet
```
dan-f: What's the use case for this being public?

dan-f:
:+1: once we figure out names!

src/js/utils/html-utils.js
andy-armstrong: Whoops! I've fixed it.

src/js/utils/html-utils.js
andy-armstrong: See @robrap's comment earlier in this PR:

{quote}As you can see in my comment on the "Implement in the platform ticket" ([UITK-75](https://openedx.atlassian.net/browse/UITK-75)), code outside of this method may want to check its type (String or HTMLSnippet), and determine whether or not it still needs to escape.{quote}

It also makes it somewhat clearer to document the helper methods as they can say what type they return.

src/js/utils/html-utils.js
```diff
@@ -0,0 +1,115 @@
...
+    'use strict';
+    define(['underscore'], function(_) {
+        var interpolate, escapeInterpolateParameters, HTML;
+
+        /**
+         * Creates an HTML snippet.
+         *
+         * The intention of an HTML snippet is to capture a string that
+         * is known to contain HTML that has been safely escaped.
+         * This allows interpolate to understand that it does not
```
robrap: Possible adjustment of "This allows interpolate to..." to "As an example, this allows interpolate to...".  This might clarify that other outside code could have the same understanding as interpolate.

src/js/utils/specs/html-utils-spec.js
```diff
@@ -0,0 +1,70 @@
+define(
+    [
+        '../../utils/spec-helpers/spec-helpers.js',
+        '../html-utils.js'
+    ],
+    function(SpecHelpers, HtmlUtils) {
+        'use strict';
+
+        describe('HTMLUtils', function() {
```
robrap: HTMLUtils => HtmlUtils

robrap:
Thanks @andy-armstrong.  This looks great to me.  I left 2 very minor comments.  :+1:  with those changes.  I'm happy with the current names.

That said, I have one possible additional enhancement.  Let's say we have the following example:
```
HtmlUtils.interpolate(
    gettext('You are enrolling a Rock & Roll star!')
)
```
This does exactly what we want, but the name `interpolate()` doesn't make a lot of sense because there is no interpolation needed.  We need to document what we'd expect to see used here.  One alternative that would work is:
```
_.escape(
    gettext('You are enrolling a Rock & Roll star!')
)
```
I'd propose we also have HtmlUtils provide the same so we can use it as an option (and as the best practice):
```
HtmlUtils.escape(
    gettext('You are enrolling a Rock & Roll star!')
)
```
I'd like your thoughts on this.  Do you agree, or would you prefer to keep `_.escape()` as the existing alternative because it is possibly well known and already documented as part of underscore?

Thanks.

src/js/utils/html-utils.js
andy-armstrong: Fixed, thanks

src/js/utils/specs/html-utils-spec.js
andy-armstrong: Fixed, and found a couple more too.

andy-armstrong:
@robrap I really liked your suggestion of providing ``HtmlUtils.escape`` as a first class function. It isolates our best practices from Underscore (which we might drop at some point). More importantly, I found that it is useful to be able to pass HtmlSnippets to ``escape`` and have them pass through without being escaped again. In fact, having this method allowed me to remove the rather ugly internal method ``escapeInterpolateParameters`` with just ``_.mapObjects(parameters, escape)``. I'm also going to be using this method in edx-platform for exactly the use cases you describe, and in some of them the code won't know whether it has a raw string or an HtmlSnippet.

@robrap @dan-f @AlasdairSwan please re-review.

andy-armstrong:
@dan-f I've been thinking through the naming options for ``HtmlUtils.HTML``. I don't think we can use ``HtmlUtils.Html`` because that naming convention should indicate that ``Html`` is a class but it isn't. For example, JSHint complains about a class being constructed without using the ``new`` operator. ``HtmlUtils.html`` would be okay from a naming convention, but I find that it doesn't stand out enough. My current leaning is to go with @robrap's original proposal of ``HTML`` both because it stands out and because it mirrors what we're doing in Python.

Let me know if this works for you or not.

andy-armstrong:
@robrap @dan-f In working on the conversion of some of edx-platform, I'm starting to see a need for a non-HTML interpolate function. There are places where the code needs to create a text string and it is not going to be used as HTML. For example:

https://github.com/edx/edx-platform/blob/master/lms/djangoapps/teams/static/teams/js/views/edit_team_members.js#L50

I started to convert these functions over to use ``HtmlUtils.interpolate`` but the code didn't read correctly because it was jarring to see mention of HTML. However, I'd like to use the same syntax for consistency.

What would you think of also having ``StringUtils.interpolate`` which is a pure interpolation function that has nothing to do with HTML? Would that be confusing because there are two similar functions, or would it be less confusing because otherwise folks are relying on ``interpolate`` in the global namespace which has a somewhat confusing syntax?

I wonder if it would be possible to share logic between the two functions too, or whether that would just be too contrived.

robrap:
@andy-armstrong Thanks for your continued work on this.  I must have felt it because I woke up earlier thinking about this and finally got up to comment about some of the same things.  Also, thank you for actually putting your method to use to help discover these issues early.  :)

It struck me, as it did you, that `_.escape()` and `HtmlUtils.escape()` would be different in that the latter would recognize an `HtmlSnippet`, and that this difference is important and useful.  In the markupsafe library we use in Python, it uses this same convention.  In other words, it has an `escape()` method that won't double-escape if it is being passed `Markup`, which is the equivalent of an `HtmlSnippet`.  So again, I'm fine with the name `escape()`.

I'm on board with almost all existing names, except I think we may need to go with `HtmlUtils.interpolateHtml()` and `StringUtils.interpolate()`.  Especially for cases where the function is put into a variable, I think we'd want people to use `interpolate` and `interpolateHtml` to mean different things.  The only way to somewhat enforce this naming convention would be through the different function names to begin with.  I guess I don't know how often this is done, and how often one would see `interpolate()` (using the old name) separate from `HtmlUtils`?

Note that the linter will help enforce some of this.  For example, in Underscore.js templates we could dictate:
```
// Since "<%-" escapes, the linter would be happy with anything here except those methods found below
<%- ... %>

// Since "<%=" does not escape, the linter would be ok with the following:
<%= HtmlUtils.interpolateHtml(...) %>
<%= interpolateHtml(...) %>
<%= HtmlUtils.escape(...) %>
<%= escape(...) %>
<%= _.escape(...) %>

// The linter would NOT be ok with the following:
<%= interpolate(...) %>
 
// If we wanted to stick with "HtmlUtils.interpolate", the linter would instead be ok with the following:
<%= HtmlUtils.interpolate(...) %>
```

So, the linter can help somewhat in an Underscore.js Template file, but it is going to have a much more difficult time in a plain old JavaScript file.  So, I'll start with a vote for `HtmlUtils.interpolateHtml()`.

andy-armstrong:
@robrap Thanks for the great suggestions. I've introduced ``StringUtils.interpolate`` and renamed the original method to ``HtmlUtils.interpolateHtml``. It turned out to be very simple to implement the latter in terms of the former too.

One thing that we should discuss is how developers should handle escaping and/or interpolation in Underscore templates. The challenge is that none of these helpers are available in the global namespace, so it isn't straightforward to access them. The simplest thing to do would be to have the client code pass ``HtmlUtils`` through to the template. However, this would only be a best practice and would require every developer to think of it. We could have them use ``require`` directly in the template, but I don't know whether RequireJS Optimizer will be able to handle that appropriately (we would have to try). @AlasdairSwan and @dan-f, do you have any suggestions?

robrap:
@andy-armstrong Thanks.  I'm open to whatever you think will work in practicality.  I know @dan-f doesn't like the idea of extending underscore, but clearly enough people wanted this that it is available through the library using [mixin](http://underscorejs.org/#mixin).

Would something like `_.HtmlUtils` or `_.edX.HtmlUtils` make sense as a simple name-spaced solution that provides easy access?  I'm just throwing out ideas.  I do want it to be relatively simple for developers to use what they need.  I assume that this is just the first of many things you are going to want to easily make available.

src/js/utils/html-utils.js
dan-f: If that's a common use case, then I think perhaps we should add it to this library.

andy-armstrong: Good point, @dan-f. I think my new method ``HtmlUtils.escape`` does exactly what @robrap wanted, in that it will only escape if necessary. I suppose you could now argue that that means that ``HtmlSnippet`` can become internal again. I'm tempted to leave it public so that the documentation is clearer, and because I could imagine there being other potential usages.

dan-f:
:+1: on `HtmlUtils.HTML` as the factory function to return html safe text.
:+1: on `HtmlUtils.escape` as a function which smartly escapes and never double-escapes.

As for `StringUtils.interpolate` - I'd argue that we don't need. 90% of the time that we interpolate a string, we are preparing a string for the user to see. So it's going to end up in HTML. Hence when we do the interpolation, we want to mark HTML-safe strings with the `HTML` factory. So your example would become:
```javascript
var HtmlUtils = require('HtmlUtils'),
    interpolate = HtmlUtils.interpolate,
    HTML = HtmlUtils.HTML;

_.each(this.model.get('membership'), function(membership) {
    dateJoined = interpolate(
        // Translators: 'date' is a placeholder for a fuzzy, relative timestamp (see: https://github.com/rmm5t/jquery-timeago)
        gettext("Joined %(date)s"),
        {date: HTML(self.dateTemplate({date: membership.date_joined}))}
    );

```

src/js/utils/html-utils.js
```diff
@@ -0,0 +1,114 @@
...
+         *
+         * @param {string} htmlString The string of HTML.
+         * @constructor
+         */
+        function HtmlSnippet(htmlString) {
+            this.text = htmlString;
+        }
+        HtmlSnippet.prototype.toString = function() {
+            return this.text;
+        };
```
dan-f: It could also be useful to implement `valueOf`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/valueOf

dan-f:
@andy-armstrong - I've started following a pattern in which string interpolation/preparation happens exclusively in JS, not in the template, which avoids this issue. I think that to me, it counts as view logic; the logic where we decide _what_ to present to the user. I like the template to be lightweight and simply determine _how_ that information is presented to the user.

andy-armstrong:
@dan-f The need for ``StringUtils.interpolate`` is when you are building up a string for some other purpose than HTML, e.g. to post it to AJAX, or to write it to the console, or any host of other reasons. It also might be an intermediate text string that will eventually be placed into HTML but might not be. I found that using ``HtmlUtils.interpolateHtml`` for such code just read as incorrect, and I think it would be functionally incorrect too if it was then used in a non-HTML context.

If it would help, you can see a number of such usages as I'm cleaning up teams to use these new functions:

https://github.com/edx/edx-platform/pull/11799/files

andy-armstrong:
@robrap @dan-f @AlasdairSwan I've pushed up another change which introduces a new ``HtmlUtils.template`` function. This wraps ``_.template`` but passes in the variable ``HtmlUtils`` so that all templates have access to these utilities. In addition, it returns an ``HtmlSnippet`` so that any consumers know that the result has already been escaped appropriately.

Please re-review at your convenience.

src/js/utils/html-utils.js
andy-armstrong: Great idea. Done.

andy-armstrong:
@dan-f I spoke with @robrap about your suggestion of doing the interpolation and escaping in JavaScript. His recommendation is that the code always escapes only when the context is known, so his preference is to escape in the template. His concern with your approach is that all Underscore templates will have to use ``<%=`` and assume that the parameters passed in were appropriately escaped. This will also make it impossible for his linter to detect XSS problems.

We decided to add ``HtmlUtils.template`` to make it easy for templates to do this. I'm going to try using it and see whether it helps or not.

andy-armstrong:
@bjacobel FYI this is the XSS PR that I'm currently working on.

AlasdairSwan:
Looks good @andy-armstrong :+1:

robrap:
Based on the discussion, the linter would be ok with `<%=` (which does not escape), when used with the following:
```
<%= HtmlUtils.interpolateHtml(...) %>
<%= HtmlUtils.escape(...) %>
<%= HtmlUtils.HTML(...) %>
<%= _.escape(...) %>
```

src/js/utils/specs/html-utils-spec.js
```diff
@@ -0,0 +1,106 @@
...
+                        },
+                        'You are enrolling in <span class="course-title">Rock &amp; Roll</span>'
+                    ]
+                }, function(template, options, expectedString) {
+                    var result = HtmlUtils.interpolateHtml(template, options);
+                    expect(result.toString()).toBe(expectedString);
+                });
+            });
+
+            describe('template', function() {
```
robrap: Can you add a test for the fact that it is returning an HtmlSnippet that should not get escaped again if wrapped with another call like HtmlUtils.escape()?

andy-armstrong: Added.

robrap:
:+1: Nice job.  Thanks!

dan-f:
@andy-armstrong and @robrap: I agree that you should always escape as close to the context/template as possible. You can still do so in JS:

```javascript
var template = _.template(templateName);
var message = interpolate(...);
this.$el.html(template({message: message}));
```

andy-armstrong:
@dan-f The problem with this is that the template has to now assume that the parameter it receives is escaped. So the template will have to use:

    <p><%= message %></p>

and that is precisely what @robrap's new linter will flag as being a potential issue. We want templates to default to safe (i.e. use ``<%-``) except in a few limited situations.

One way to combine your approach and still be safe is to use ``HtmlUtils.interpolateHtml`` in the JavaScript and then do the following in the template:

    <p><%= HtmlUtils.escape(message) %></p>

This template will then escape all raw strings, but will allow previously escaped HtmlSnippets to come through as is. This also won't cause an exception with the linter as ``HtmlUtils.escape`` is one of the known-to-be-safe patterns. Does that seem like a good compromise?

@robrap and I are going to be updating the best practice documentation to make all of this explicit for next week's XSS marathon.

src/js/utils/string-utils.js
```diff
@@ -0,0 +1,59 @@
...
+ */
+(function(define) {
+    'use strict';
+    define(['underscore', './string-utils.js'], function(_, StringUtils) {
+        var interpolate;
+
+        /**
+         * Returns a string created by interpolating the provided parameters.
+         *
+         * The HTML text is provided as a tokenized format string where parameters
```
cahrens: @andy-armstrong it looks like you forgot to update the documentation after this method moved out of HTMLUtils? This isn't HTML text, it's just a string, correct?

src/js/utils/string-utils.js
```diff
@@ -0,0 +1,59 @@
...
+         * are indicated via curly braces, e.g. 'Hello {name}'. These tokens are
+         * replaced by the parameter value of the same name.
+         *
+         * Parameter values will be rendered using their toString methods and then
+         * HTML-escaped. The only exception is that instances of the class HTML
```
cahrens: Is this HTML-escaping at all? I thought this method is just about interpolation.

src/js/utils/html-utils.js
```diff
@@ -0,0 +1,145 @@
...
+                _.mapObject(parameters, escape)
+            );
+            return HTML(result);
+        };
+
+        /**
+         * Returns a function that renders an Underscore template as an HTML snippet.
+         *
+         * Note: This helper function makes the following context parameters
+         * available to the template in addition to those passed in:
```
cahrens: Are there supposed to be multiple parameters? I'm having trouble understanding the documentation.

PR #31, opened 2016-03-11
Alasdair/fedx 36 dropdown menu added arrow support

AlasdairSwan:
@andy-armstrong please review. This PR adds the left and right arrow support to scroll through the dropdown list (as requested 2 hackathons ago) and it adds ```space bar``` close and ```tab+shift``` capture as per Christine's PR (https://github.com/edx/edx-platform/pull/11393)

### DIFF
--- a/src/js/dropdown-menu/dropdown-menu-view.js
+++ b/src/js/dropdown-menu/dropdown-menu-view.js
@@ -9,7 +9,7 @@ define([
         'use strict';
 
         /**
-         *  Dropdown Menu View which is fully accessible
+         *  To render a Dropdown Menu View which is fully accessible
          *
          *  Initialize the view by passing in the following attributes
          *  className: 'space separated string of classes for element',
@@ -51,8 +51,13 @@ define([
             menu: '.dropdown-menu',
 
             initialize: function(options) {
-                this.$parent = $(options.parent);
-                this.render();
+                if (options.parent) {
+                    this.$parent = $(options.parent);
+                }
+
+                this.keyBack = [constants.keyCodes.up, constants.keyCodes.left];
+                this.keyForward = [constants.keyCodes.down, constants.keyCodes.right];
+                this.keyClose = [constants.keyCodes.esc, constants.keyCodes.space];
             },
 
             className: function() {
@@ -68,10 +73,9 @@ define([
             },
 
             postRender: function() {
-                this.$menu = this.$el.find('.dropdown-menu');
+                this.$menu = this.$('.dropdown-menu');
                 this.$page = $(document);
-                this.$dropdownButton = $('.js-dropdown-button');
-                this.listenForPageKeypress();
+                this.$dropdownButton = this.$('.js-dropdown-button');
             },
 
             /**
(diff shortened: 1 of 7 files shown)


AlasdairSwan:
@dsjen if you have a minute could you please review this PR?

src/js/dropdown-menu/dropdown-menu-view.js
```diff
@@ -142,24 +142,32 @@ define([
...
             },
 
+
+            focusLastItem: function() {
+                this.$menu.find('.action.last').focus();
+            },
+
             handlerIsAction: function(key, $el) {
-                if (key === constants.keyCodes.up) {
+                if (key === constants.keyCodes.up || key === constants.keyCodes.left) {
```
dsjen: This is just a suggestion.  Have you thought about using underscore's [contain](http://underscorejs.org/#contains)? e.g. `_([constants.keyCodes.up, constants.keyCodes.left]).contains(key)`.  Maybe even creating member variables and using them::

```
previewKeys = [constants.keyCodes.up, constants.keyCodes.left];
...
if (_(this.previousKeys).contains(key) {...}
```

It may be a bit overkill for two keys though!

src/js/dropdown-menu/dropdown-menu-view.js
```diff
@@ -142,24 +141,32 @@ define([
                 this.$menu.find('.action').first().focus();
             },
 
+
+            focusLastItem: function() {
+                this.$menu.find('.action.last').focus();
```
dsjen: What about `this.$menu.find('.action').last().focus()`?  I'm not sure if the first item has a `first` class and it seems a bit inconsistent to do it differently here.  Just wondering!

dsjen:
I left two questions that are just "nits."  I think most code paths will be hit with the current tests.  My last suggestion is to add a test that hits the `focusLastItem()`--up to you though.

LGTM! :+1:

src/js/dropdown-menu/dropdown-menu-view.js
```diff
@@ -42,6 +42,8 @@ define([
         var DropdownMenuView = Backbone.View.extend({
             tpl: _.template(DropdownTpl),
 
+            // For event handling only instances do I need to scope this somehow?
```
andy-armstrong: These events will only be scoped to the root element of the view:

http://backbonejs.org/#View-events

I think that's what you want. If so, you can just remove this comment.

src/js/dropdown-menu/dropdown-menu-view.js
```diff
@@ -51,8 +53,13 @@ define([
             menu: '.dropdown-menu',
 
             initialize: function(options) {
-                this.$parent = $(options.parent);
-                this.render();
+                if ( options.parent ) {
```
andy-armstrong: Nit: we agreed to not put spaces around the parenthesis (sorry!)

src/js/dropdown-menu/dropdown-menu-view.js
```diff
@@ -130,9 +136,9 @@ define([
             },
 
             escKeypressHandler: function(event) {
-                var keyCode = event.keyCode;
+                var key = event.keyCode;
 
-                if (keyCode === constants.keyCodes.esc) {
+                if (_.contains(this.keyClose, key)) {
                     // When the ESC key is pressed, close all menus
                     this.closeDropdownMenus(true);
```
andy-armstrong: Why is this plural? Won't each dropdown have only a single menu? Is it closing all menus on the page? I'm not a big fan of affecting other views from within this view. Pressing escape won't close other modals, so why would it understand other dropdowns?

src/js/dropdown-menu/specs/dropdown-menu-view-spec.js
```diff
@@ -1,9 +1,11 @@
 define([
         'jquery',
         'underscore',
-        '../dropdown-menu-view.js'
+        '../dropdown-menu-view.js',
+        '../../utils/constants.js',
+        '../../../../test/jquery.simulate.js'
```
andy-armstrong: Nit: I'd refer to this via an absolute path. IMO relative paths only make sense if there is a relationship between the two files. This relative path would break if we happened to move the dropdown menu code for some reason.

andy-armstrong:
Some minor nits from me, but looks great overall. :+1:

src/js/dropdown-menu/dropdown-menu-view.js
AlasdairSwan: Yes, this currently closes all dropdowns but I see your point about scope. I will update to only close the current view's menu.

PR #34, opened 2016-03-16
Improve HTML snippet handling for edx-platform

andy-armstrong:
It turns out that JQuery doesn't call ``toString`` on objects passed to its methods, so our approach of wrapping an HTML string in an object didn't work. At @robrap's suggestion, I'm now creating a string object and setting an ``isHtmlSnippet`` attribute on it. Fortunately almost all of the tests just worked with this fix, and even better you don't need to call ``toString`` before comparing the results.

In addition, I've removed a relative dependency from ``HtmlUtils`` to ``StringUtils`` as that doesn't currently work in edx-platform's Jasmine tests (I'm hoping it will work once we switch to Karma). In order to do this, I've changed the Jasmine tests to proxy dependency paths starting with ``edx-ui-toolkit`` to the correct location.

I'm successfully using these changes in my PR to incorporate the UI Toolkit into the platform:

https://github.com/edx/edx-platform/pull/11799

@robrap @dan-f please review.

FYI @AlasdairSwan @bjacobel

### DIFF
--- a/src/js/utils/html-utils.js
+++ b/src/js/utils/html-utils.js
@@ -6,7 +6,7 @@
  */
 (function(define) {
     'use strict';
-    define(['underscore', './string-utils.js'], function(_, StringUtils) {
+    define(['underscore', 'edx-ui-toolkit/js/utils/string-utils'], function(_, StringUtils) {
         var escape, interpolateHtml, HTML, template;
 
         /**
--- a/src/js/utils/html-utils.js
+++ b/src/js/utils/html-utils.js
@@ -10,54 +10,33 @@
         var escape, interpolateHtml, HTML, template;
 
         /**
-         * Creates an HTML snippet.
-         *
-         * The intention of an HTML snippet is to capture a string that
-         * is known to contain HTML that has been safely escaped.
-         * As an example, this allows interpolate to understand that
-         * it does not need to further escape this HTML.
+         * Helper function to return a string flagged as an HTML snippet.
          *
          * @param {string} htmlString The string of HTML.
-         * @constructor
-         */
-        function HtmlSnippet(htmlString) {
-            this.text = htmlString;
-        }
-        HtmlSnippet.prototype.valueOf = function() {
-            return this.text;
-        };
-        HtmlSnippet.prototype.toString = function() {
-            return this.text;
-        };
-
-        /**
-         * Helper function to create an HTML snippet.
-         *
-         * @param {string} htmlString The string of HTML.
-         * @returns {HtmlSnippet} An HTML snippet that can be safely rendered.
+         * @returns {string} A string flagged as an HTML snippet.
          * @constructor
          */
         HTML = function(htmlString) {
-            return new HtmlSnippet(htmlString);
+            var htmlSnippet = new String(htmlString); // jshint ignore:line
+            htmlSnippet.isHtmlSnippet = true;
+            return htmlSnippet;
         };
 
         /**
-         * Returns HTML that is appropriately escaped.
+         * Returns an HTML string that is appropriately escaped.
          *
-         * If the input is provided as a string, then an HtmlSnippet
-         * is returned with the original text escaped. If the input
-         * is already an HtmlSnippet then it is known to be already
-         * escaped and so it is just returned directly.
+         * Note: if the string provided is flagged as being an HTML snippet
+         * then it will be returned directly so as not to double escape it.
          *
-         * @param {(string|HtmlSnippet)} text Some HTML that is
-         * either provided as a string or as an HtmlSnippet.
-         * @returns {HtmlSnippet} A safely escaped HtmlSnippet.
+         * @param {string} text A string that is either plan text, or is
+         * flagged as being an HTML snippet.
+         * @returns {string} A safely HTML-escaped string.
          */
         escape = function(text) {
-            if (text instanceof HtmlSnippet) {
-                return text;
+            if (text.isHtmlSnippet) {
+                return text.toString();
             } else {
-                return HTML(_.escape(text));
+                return _.escape(text);
             }
         };
 
@@ -97,7 +76,8 @@
          *
          * @param {string} formatString The string to be interpolated.
          * @param {Object} parameters An optional set of parameters to the template.
-         * @returns {HtmlSnippet} The resulting safely escaped HTML snippet.
+         * @returns {string} The resulting safely escaped string flagged as
+         * an HTML snippet.
          */
         interpolateHtml = function(formatString, parameters) {
             var result = StringUtils.interpolate(
@@ -108,7 +88,7 @@
         };
 
         /**
-         * Returns a function that renders an Underscore template as an HTML snippet.
+         * Returns a function that renders an Underscore template as a string.
          *
          * Note: This helper function makes the following context parameters
          * available to the template in addition to those passed in:
@@ -117,7 +97,7 @@
          *
          * @param {string} text
          * @param {object} settings
-         * @returns {function} A function that returns a rendered HTML snippet.
+         * @returns {function} A function that returns a rendered string.
          */
         template = function(text, settings) {
             var HtmlUtils = this,
@@ -129,7 +109,7 @@
                         },
                         data
                     );
-                    return HTML(rawTemplate(augmentedData));
+                    return rawTemplate(augmentedData);
                 };
             return template;
         };
@@ -137,7 +117,6 @@
         return {
             escape: escape,
             HTML: HTML,
-            HtmlSnippet: HtmlSnippet,
             interpolateHtml: interpolateHtml,
             template: template
         };
--- a/src/js/utils/specs/html-utils-spec.js
+++ b/src/js/utils/specs/html-utils-spec.js
@@ -1,21 +1,26 @@
 define(
     [
+        'jquery',
         '../../utils/spec-helpers/spec-helpers.js',
         '../html-utils.js'
     ],
-    function(SpecHelpers, HtmlUtils) {
+    function($, SpecHelpers, HtmlUtils) {
         'use strict';
 
         describe('HtmlUtils', function() {
-            describe('HtmlSnippet', function() {
-                it('can convert to a string', function() {
-                    expect(new HtmlUtils.HtmlSnippet('Hello, world').toString()).toBe('Hello, world');
-                });
-            });
-
             describe('HTML', function() {
-                it('returns an HtmlSnippet', function() {
-                    expect(HtmlUtils.HTML('Hello, world') instanceof HtmlUtils.HtmlSnippet).toBeTruthy();
+                it('returns a string flagged as an HTML snippet', function() {
+                    expect(HtmlUtils.HTML('Hello, world').isHtmlSnippet).toBeTruthy();
+                });
+
+                it('returns an HTML snippet that can be used by JQuery', function() {
+                    var testHtmlString = '<p>A test</p>',
+                        htmlSnippet = HtmlUtils.HTML(testHtmlString);
+                    setFixtures('<div class="test"></div>');
+                    $('.test').html(htmlSnippet);
+                    expect($('.test').html()).toBe(testHtmlString);
+                    $('.test').append(htmlSnippet);
+                    expect($('.test').html()).toBe(testHtmlString + testHtmlString);
                 });
             });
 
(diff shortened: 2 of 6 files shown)


edx-webhook:
## PR Cover Letter

## Sandbox

## Testing
- [ ] i18n
- [ ] RTL
- [ ] Analytics
- [ ] Performance

## Reviewers
- [ ] Code review #1
- [ ] Code review #2
- [ ] UX review
- [ ] Accessibility review
- [ ] Product review

## DevOps assistance
- configuration changes
- feature flags
- migrations
- [ ] File a support ticket

## Post-review
- [ ] Squash commits

andy-armstrong:
@robrap @dan-f @AlasdairSwan @bjacobel Well, it turns out that JQuery doesn't always behave nicely with ``new String('...')`` instances either. Sometimes we get errors like this:

    TypeError: Failed to execute 'appendChild' on 'Node': parameter 1 is not of type 'Node'.

It seems that the code has done a test to see if the parameter is a primitive string, and if not then it assumes that it is a node. I don't see any option right now other than sprinkling ``toString()`` everywhere, which seems brittle and hard to debug.

I'm tempted to say that ``HtmlUtils.template`` and ``HtmlUtils.escape`` should return primitive strings, which means that they can't safely be chained with other calls. I'm not sure what to do about ``HtmlUtils.intepolateHtml`` because it does seem quite likely to chain the result of this into another HTML snippet.

Thoughts, anyone?

robrap:
@andy-armstrong and I discussed, and he has a couple more ideas we will be trying.  Stay tuned...

andy-armstrong:
Here's where @robrap and I landed after discussions today:

Given how JQuery special cases primitive strings, we decided to change ``ViewUtils.escape`` and ``ViewUtils.template`` to return primitive strings rather than HTML snippets, since the HTML output of these functions is almost always passed on to JQuery. The only downside to this is that you cannot directly chain results together, e.g. use the result of rendering a template as a parameter to another template. Instead you need to explicitly wrap the result again using ``ViewUtils.HTML`` which, while a little redundant, is at least very clear as to what you are doing.

The more challenging aspect is ``ViewUtils.interpolateHtml``. This function is very much used for creating partial HTML snippets, so we decided to leave its behavior as is. This works well for most use cases, but it does not always work with JQuery. Our recommendation is to use ``toString`` to convert the snippet into a primitive string in this situation. 

I wrote some tests in the UI Toolkit to demonstrate the JQuery problem with string objects, but they don't fail which leads me to suspect that this issue was addressed in a JQuery update. I also searched the repository and was unable to find the error messages in question. Our hope is that when we upgrade our JQuery version that these explicit workarounds won't be necessary.

@robrap @dan-f please review.

dan-f:
:+1: on having developers explicitly call `toString` on `HtmlSnippet`s.

dan-f:
@andy-armstrong - actually new idea.

jQuery lets you pass a function returning a string $.fn.html and $.fn.text.

So what if we were to do something like:

```javascript
var HtmlSnippet = function (str) {
    var ret = function () { return str };
    Object.setPrototypeOf(ret, HtmlSnippet.prototype);
    return ret;
}

var snip = new HtmlSnippet('foo');
snip instanceof HtmlSnippet; // -> true
snip(); // -> 'foo'
```

src/js/utils/html-utils.js
```diff
@@ -6,58 +6,40 @@
...
-        }
-        HtmlSnippet.prototype.valueOf = function() {
-            return this.text;
-        };
-        HtmlSnippet.prototype.toString = function() {
-            return this.text;
-        };
-
-        /**
          * Helper function to create an HTML snippet.
```
robrap: Comment is out of date.

src/js/utils/html-utils.js
```diff
@@ -6,58 +6,40 @@
...
-         * either provided as a string or as an HtmlSnippet.
-         * @returns {HtmlSnippet} A safely escaped HtmlSnippet.
+         * @param {string} text Some HTML that is either a raw string
+         * or one that is flagged as an HTML snippet.
+         * @returns {string} A safely HTML-escaped string.
          */
         escape = function(text) {
-            if (text instanceof HtmlSnippet) {
-                return text;
+            if (text.isHtmlSnippet) {
```
robrap: FYI: Feel free to ignore this comment.  I simply documented a doubt, followed by how I moved past it. :)

Would the outside world ever want to know if something is an HtmlSnippet?  Do we want an `HtmlUtils.isHtmlSnippet()` call for completeness and to encapsulate how this is done?

After thinking a bit more about this, I guess even if we one day decided to return something of type HtmlSnippet, it could still have an isHtmlSnippet property, so this should be fine.

src/js/utils/html-utils.js
```diff
@@ -129,7 +111,7 @@
                         },
                         data
                     );
-                    return HTML(rawTemplate(augmentedData));
+                    return rawTemplate(augmentedData);
```
robrap: @andy-armstrong Your happier with this?  I thought someone needed to use `toString()` and the jQuery upgrade would improve this and allow for backward compatibility?  If you return a String here, you must always return a String here, and we wouldn't be able to improve it unless we added a new method like `templateAsHtmlSnippet`, which gets long enough that we may always have to require using the wrapper.

Would you mind determine if the jQuery upgrade really does influence this before we commit one way or another?

src/js/utils/html-utils.js
```diff
@@ -6,58 +6,40 @@
...
-        HtmlSnippet.prototype.toString = function() {
-            return this.text;
-        };
-
-        /**
          * Helper function to create an HTML snippet.
          *
          * @param {string} htmlString The string of HTML.
-         * @returns {HtmlSnippet} An HTML snippet that can be safely rendered.
+         * @returns {string} An string flagged as an HTML snippet.
```
robrap: An string => A string

src/js/utils/specs/html-utils-spec.js
```diff
@@ -1,21 +1,26 @@
...
-    function(SpecHelpers, HtmlUtils) {
+    function($, SpecHelpers, HtmlUtils) {
         'use strict';
 
         describe('HtmlUtils', function() {
-            describe('HtmlSnippet', function() {
-                it('can convert to a string', function() {
-                    expect(new HtmlUtils.HtmlSnippet('Hello, world').toString()).toBe('Hello, world');
+            describe('HTML', function() {
+                it('returns an string flagged as an HTML snippet', function() {
```
robrap: an string => a string

robrap:
@dan-f Andy and I talked about a solution like what you mentioned, only not as elegant.  However, I think the issue is that this would make the jQuery methods happier, but it doesn't work well in the many places where we want either a String or an object with `.toString()`, like in the templates.  

So, then the question becomes whether or not you introduce separate functions for jQuery usage specifically, but this started to get confusing and basically calling `toString()` on the returned value ends up giving you exactly this.

As @andy-armstrong mentioned, the best practices may get simpler (no need to call toString()) once we get the jQuery upgrade, and it will be backward compatible (i.e. it would fine to leave toString() even if you didn't need it).  @andy-armstrong Can you let us know when you confirm this?

@andy-armstrong I assume you meant `HtmlUtils` instead of `ViewUtils`.

src/js/utils/html-utils.js
andy-armstrong: This is what we discussed, which is that templates are very rarely (I don't even know of a case) evaluated and then passed on to another interpolation function where the fact that they are an HTML snippet becomes important. I think I'm okay with saying that folks should do the following if they want to pass a template result to someone else:

    HtmlUtils.interpolateHtml('<div><h2>Inline template</h2>{template_html}</div>',
        {
            template_html: HtmlUtils.HTML(HtmlUtils.template(...))
        }
    );

If we don't do this now then we'll have to update every ``HtmlUtils.template`` call to also call ``toString`` which seems error prone to me. If the JQuery upgrade fixes the issue then folks can remove the redundant ``HtmlUtils.HTML`` wrapper at that time.

andy-armstrong:
@robrap Yes, I don't know where ``ViewUtils`` came from. :-)

andy-armstrong:
@robrap I've fixed the two doc typos, and squashed the commits. Hopefully this is now in a state where we can merge. Thanks!

src/js/utils/html-utils.js
```diff
@@ -6,58 +6,40 @@
...
          */
         HTML = function(htmlString) {
-            return new HtmlSnippet(htmlString);
+            var htmlSnippet = new String(htmlString); // jshint ignore:line
+            htmlSnippet.isHtmlSnippet = true;
+            return htmlSnippet;
         };
 
         /**
          * Returns HTML that is appropriately escaped.
```
robrap: Returns an HTML string that is appropriately escaped.

src/js/utils/html-utils.js
```diff
@@ -6,58 +6,40 @@
...
          *
-         * If the input is provided as a string, then an HtmlSnippet
-         * is returned with the original text escaped. If the input
-         * is already an HtmlSnippet then it is known to be already
-         * escaped and so it is just returned directly.
+         * If the input is provided as a string, then a new string
+         * is returned with the original text escaped. In addition,
+         * the string is flagged as being an HTML snippet. If the input
```
robrap: Fix comment.  This is not an HTML Snippet.  You may want to provide a sample of wrapping with HTML() if they need a snippet.

src/js/utils/html-utils.js
```diff
@@ -6,58 +6,40 @@
...
+         * is already a string that is an HTML snippet then it is known
+         * to be already escaped and so it is just returned directly.
          *
-         * @param {(string|HtmlSnippet)} text Some HTML that is
-         * either provided as a string or as an HtmlSnippet.
-         * @returns {HtmlSnippet} A safely escaped HtmlSnippet.
+         * @param {string} text Some HTML that is either a raw string
```
robrap: "Some HTML" is not really accurate.
Does this work for you:
A string that is either plain text, or is flagged as an HTML snippet.

src/js/utils/html-utils.js
```diff
@@ -97,7 +79,7 @@
          *
          * @param {string} formatString The string to be interpolated.
          * @param {Object} parameters An optional set of parameters to the template.
-         * @returns {HtmlSnippet} The resulting safely escaped HTML snippet.
+         * @returns {string} The resulting safely escaped HTML snippet.
```
robrap: Do you want to define somewhere that an "HTML snippet" is a "string flagged as an HTML snippet"?  Or, would it be better to consistently use the latter through all comments?
For example:
The resulting safely escaped string flagged as an HTML snippet.

src/js/utils/html-utils.js
robrap: If we want to be consistent, this should read:
Helper function to convert to a string flagged as an HTML snippet.

src/js/utils/html-utils.js
```diff
@@ -108,7 +90,7 @@
         };
 
         /**
-         * Returns a function that renders an Underscore template as an HTML snippet.
+         * Returns a function that renders an Underscore template as an HTML string.
```
robrap: Maybe just drop the word "HTML"?  It really all depends on what is done in the template.

src/js/utils/html-utils.js
```diff
@@ -117,7 +99,7 @@
          *
          * @param {string} text
          * @param {object} settings
-         * @returns {function} A function that returns a rendered HTML snippet.
+         * @returns {function} A function that returns a rendered HTML string.
```
robrap: Again, drop "HTML"?

src/js/utils/html-utils.js
robrap: I'm all set.  Thanks.

robrap:
:+1: After comment clean up.  Let me know if there is anything specific you want me to look at.  Thanks.

dan-f:
I don't think a jQuery upgrade will make the `new String('foo')` pattern work: https://jsfiddle.net/o188evht/

robrap:
Thanks @dan-f.  Looks like @andy-armstrong's current solution is the right one and will stay the right one. :)

andy-armstrong:
@dan-f You had a typo in your fiddle... I've fixed it and it does now work as far back as JQuery 1.6.

dan-f:
whoops! you're right. I was experimenting with passing an object with a toString method, but forgot to delete the curlies.

PR #36, opened 2016-03-16
Trap tab nav within the dropdown

AlasdairSwan:
@andy-armstrong @dsjen can I please get a review of this minor a11y change to trap tab navigation within the open menu. Arrow nav was already trapped but an update to remove the .last class from the last li introduced a regression that stopped the tabs being trapped. @clytwynec signed off on the fix in the https://github.com/edx/ux-pattern-library/pull/305 PR.

### DIFF
--- a/src/js/dropdown-menu/dropdown-menu-view.js
+++ b/src/js/dropdown-menu/dropdown-menu-view.js
@@ -76,6 +76,7 @@ define([
                 this.$menu = this.$('.dropdown-menu');
                 this.$page = $(document);
                 this.$dropdownButton = this.$('.js-dropdown-button');
+                this.$lastItem = this.$menu.find('li:last-child a');
             },
 
             /**
@@ -125,11 +126,11 @@ define([
             },
 
             focusFirstItem: function() {
-                this.$menu.find('.action').first().focus();
+                this.$menu.find('.dropdown-item:first-child .action').focus();
             },
 
             focusLastItem: function() {
-                this.$menu.find('.action').last().focus();
+                this.$lastItem.focus();
             },
 
             handlerIsAction: function(key, $el) {
@@ -145,6 +146,7 @@ define([
                     this.focusFirstItem();
                 // if up arrow or left arrow key pressed or shift+tab
                 } else if (_.contains(this.keyBack, key) || key === constants.keyCodes.tab && event.shiftKey) {
+                    event.preventDefault();
                     this.focusLastItem(); 
                 }
             },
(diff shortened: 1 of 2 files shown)


andy-armstrong:
:+1: Looks good.

It would be ideal if you could add a test for this, but I know that focus tests are a pain to write. Having a test would stop this kind of regression happening in the future. Bonus points if you can add such a test.

src/js/dropdown-menu/dropdown-menu-view.js
```diff
@@ -76,6 +76,7 @@ define([
                 this.$menu = this.$('.dropdown-menu');
                 this.$page = $(document);
                 this.$dropdownButton = this.$('.js-dropdown-button');
+                this.$lastItem = this.$menu.find('li:last-child a');
```
dsjen: Out of curiosity, did `this.$menu.find('.action').last()` not work out?

AlasdairSwan: There was an issue when I wasn't calling event.preventDefault() on a tab + shift that made it jump to the penultimate ```li``` and I didn't realize this was the issue for a while. I do think the current way could be a little more performant 'tho because only creating the jQuery element once as opposed to every time it is needed.

src/js/dropdown-menu/dropdown-menu-view.js
```diff
@@ -227,7 +229,7 @@ define([
                     event.preventDefault();
                 }
 
-                if (key === constants.keyCodes.tab && $el.hasClass('last')) {
+                if (key === constants.keyCodes.tab && !event.shiftKey && $el[0] === this.$lastItem[0]) {
```
dsjen: Nit: Is it at all possible that these might be null/undefined?  Maybe http://underscorejs.org/#first would be safer?

dsjen:
:+1: I had one nitty question about perhaps using `_.first()`.

AlasdairSwan:
@andy-armstrong I do have some focus tests and have added a few more but I am having difficulty faking tab keypress events. I think having the tests is necessary 'tho because we really need these elements to be fully tested. We discussed moving to an npm keypress fake library which may solve the issues I am having so I think I will take a look into whether that will help.

AlasdairSwan:
@andy-armstrong after spending some time trying to get tab keypress fakes and focus tests working I think I am going to punt on the bonus points and create a new story to update the keypress library to a node based one and add the requirement to include tab focus tests there. If you're fine with that I will merge this PR and start a new one to address the UX feedback from the pattern library PR.

andy-armstrong:
Works for me @AlasdairSwan. :+1:

PR #38, opened 2016-03-18
Unique id for each dropdown menu and using UXPL styles where possible

AlasdairSwan:
@andy-armstrong @dj this PR enables an icon to be defined in the model which will render one of the icons from the UXPL (currently the assumption is being made that edx-icons.js is already loaded on the page. It will be a larger discussion to determine whether we want to add that into the UI Toolkit or not). It also updated the template file to use styles defined by the UXPL.

### DIFF
--- a/src/js/dropdown-menu/dropdown-menu-view.js
+++ b/src/js/dropdown-menu/dropdown-menu-view.js
@@ -20,7 +20,10 @@ define([
          *          text: 'username',
          *          url: 'dashboard'
          *      },
-         *      button_label: 'User options dropdown',
+         *      button: {
+         *          icon: 'icon-angle-down',
+         *          label: 'User options dropdown'
+         *      },
          *      items: [
          *          {
          *              text: 'Dashboard',
@@ -48,6 +51,8 @@ define([
                 'keydown': 'viewKeypress'
             },
 
+            dropdownButton: '.js-dropdown-button',
+
             menu: '.dropdown-menu',
 
             initialize: function(options) {
@@ -55,6 +60,7 @@ define([
                     this.$parent = $(options.parent);
                 }
 
+                this.menuId = options.menuId || 'dropdown-menu-' + this.cid;
                 this.keyBack = [constants.keyCodes.up, constants.keyCodes.left];
                 this.keyForward = [constants.keyCodes.down, constants.keyCodes.right];
                 this.keyClose = [constants.keyCodes.esc, constants.keyCodes.space];
@@ -65,6 +71,12 @@ define([
             },
 
             render: function() {
+                /**
+                 * Set in the render function to prevent error when
+                 * view is used with a pre-rendered DOM
+                 */
+                this.model.set({menuId: this.menuId});
+
                 this.$el.html(this.tpl(this.model.toJSON()));
                 this.$parent.replaceWith(this.$el);
                 this.postRender();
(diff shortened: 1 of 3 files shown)


src/js/dropdown-menu/dropdown-menu-view.js
```diff
@@ -102,7 +114,22 @@ define([
             },
 
             clickCloseDropdown: function(event, context) {
-                var $el = $(event.target);
+                var $el = $(event.target),
+                    $btn;
+
+                /** When using UXPL icons the target is sometimes
```
andy-armstrong: Nit: we don't call it UXPL anymore :-)

src/js/dropdown-menu/dropdown-menu-view.js
```diff
@@ -102,7 +114,22 @@ define([
             },
 
             clickCloseDropdown: function(event, context) {
-                var $el = $(event.target);
+                var $el = $(event.target),
```
andy-armstrong: You probably want to use ``$(event.currentTarget)`` to always get the object that you assigned the listener to:

http://stackoverflow.com/questions/5921413/difference-between-e-target-and-e-currenttarget

src/js/dropdown-menu/dropdown.underscore
```diff
@@ -7,12 +7,19 @@
     <% } %>
     <%- main.text %>
 </a>
-<button type="button" class="button button-more has-dropdown js-dropdown-button" aria-haspopup="true" aria-expanded="false" aria-controls="edx-user-menu">
-    <span class="sr-only"><%- button_label %></span>
+<button type="button" class="menu-button button-more has-dropdown js-dropdown-button <% if (!button.icon) { %>default-icon<% } %>" aria-haspopup="true" aria-expanded="false" aria-controls="<%- menuId %>">
+    <% if (button.icon) { %>
+        <span class="icon-fallback icon-fallback-img">
+            <span class="icon <%- button.icon %>" aria-hidden="true"></span>
+            <span class="text"><%- button.label %></span>
```
andy-armstrong: Why is the label ``sr-only`` if the button doesn't have an icon, but shown to everyone if it does?

src/js/dropdown-menu/specs/dropdown-menu-view-spec.js
```diff
@@ -102,7 +102,9 @@ define([
                         text: 'username',
                         url: 'dashboard'
                     },
-                    button_label: 'User options dropdown',
+                    button: {
+                        label: 'User options dropdown'
```
andy-armstrong: It would be good to add some tests for a dropdown with an icon. In particular, test that clicking on the icon has the same behavior as clicking on the button.

andy-armstrong:
This is a great addition. I have a few minor comments so let me know when I should look at it again.

It would be good to also have a11y review on this.

src/js/dropdown-menu/dropdown.underscore
AlasdairSwan: Because a pseudo element generates a caret for non-icons. The button label is still hidden for line 14, but using the icon syntax from the pattern library.

AlasdairSwan:
re. an a11y review, Christine has already done one on the pattern library branch. Which highlights the weirdness of this workflow. Maybe we can discuss in standup?

src/js/dropdown-menu/dropdown-menu-view.js
AlasdairSwan: Actually it is the target I want, not the currentTarget (the listener is on the page level and I am checking whether the user clicked on the dropdown button or not).

AlasdairSwan:
@andy-armstrong updated

src/js/dropdown-menu/dropdown-menu-view.js
andy-armstrong: I thought from the way the event was registered, that it was for items on the page with the class ``dropdown``. Can you double check that ``currentTarget`` isn't what you want, because it would simplify this logic.

andy-armstrong:
:+1: Looks great.

PR #45, opened 2016-03-25
Benp/fixup uitk16

benpatterson:
WIP Attempt to continue Alasdair's work. New PR so that it goes against master and runs travis.

### DIFF
--- a/gulp/tasks/coverage.js
+++ b/gulp/tasks/coverage.js
@@ -1,0 +1,11 @@
+(function() {
+    'use strict';
+
+    var gulp = require('gulp'),
+        coveralls = require('gulp-coveralls');
+
+    gulp.task('coverage', function() {
+        return gulp.src('coverage/**/lcov.info')
+            .pipe(coveralls());
+    });
+}());
--- a/test/karma.ci.conf.js
+++ b/test/karma.ci.conf.js
@@ -17,8 +17,8 @@ module.exports = function(config) {
         browsers: ['Chrome_travis_ci', 'Firefox', 'PhantomJS'],
         logLevel: config.LOG_INFO,
         preprocessors: {
-            'components/**/*.js': ['coverage']
+            'src/js/**/*.js': ['coverage']
         },
-        reporters: ['spec', 'coverage']
+        reporters: ['spec', 'coverage', 'coveralls']
     });
 };
--- a/gulp/tasks/default.js
+++ b/gulp/tasks/default.js
@@ -1,5 +1,5 @@
 (function() {
     'use strict';
     var gulp = require('gulp');
-    gulp.task('default', ['test', 'lint', 'jscs']);
+    gulp.task('default', ['test', 'lint', 'jscs', 'coverage']);
 }());
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
   },
   "devDependencies": {
     "gulp": "~3.9.1",
+    "gulp-coveralls": "~0.1.4",
     "gulp-jscs": "~3.0.2",
     "gulp-jshint": "~2.0.0",
     "gulp-karma": "~0.0.5",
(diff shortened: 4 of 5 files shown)


benpatterson:
@AlasdairSwan looks like we made it!

AlasdairSwan:
Awesome @benpatterson, thanks for following up on this.

AlasdairSwan:
@benpatterson I added a commit to limit the coverage to not include test and test helper files, as well as limiting the coverage run to Travis CI builds.

AlasdairSwan:
@andy-armstrong @dsjen please review

gulp/tasks/coverage.js
```diff
@@ -0,0 +1,14 @@
...
+    'use strict';
+
+    var gulp = require('gulp'),
+        coveralls = require('gulp-coveralls');
+
+    gulp.task('coverage', function() {
+        // Only run on Travis
+        if (process.argv.indexOf('--ci') !== -1) {
+            return gulp.src('coverage/**/lcov.info')
+  .             pipe(coveralls());
```
dsjen: The `.` seems like it should go at the end of the line above.

gulp/tasks/coverage.js
```diff
@@ -0,0 +1,14 @@
+(function() {
+    'use strict';
+
+    var gulp = require('gulp'),
+        coveralls = require('gulp-coveralls');
+
+    gulp.task('coverage', function() {
+        // Only run on Travis
+        if (process.argv.indexOf('--ci') !== -1) {
```
dsjen: What happens if this is run locally?  If it fails, a check to see if process exists might be helpful--maybe even with a warning message.

benpatterson:
Sounds good @AlasdairSwan. Feel free to squash over my commit when the time comes to merge. Note that it'll report coverage for `gulp test` and for `gulp test --ci` but if you don't have the env variable set, I'm pretty sure it just skips coveralls.

Changes LGTM though I'll defer to Dennis and Andy as well.

gulp/tasks/coverage.js
AlasdairSwan: @dsjen see Ben's comment below. It looks like I could probably just remove the if statement entirely. Does that seem fine?

dsjen: Sure! :cake:

dsjen:
Very exciting! :+1:

test/karma.ci.conf.js
```diff
@@ -17,8 +17,8 @@ module.exports = function(config) {
         browsers: ['Chrome_travis_ci', 'Firefox', 'PhantomJS'],
         logLevel: config.LOG_INFO,
         preprocessors: {
-            'components/**/*.js': ['coverage']
+            'src/js/**/!(*spec|*helpers).js': ['coverage']
```
andy-armstrong: I seem to remember that we keep coverage on test files in edx-platform so that we can be sure that all the tests are being run. Also, the helpers are actually part of the public API for the UI Toolkit, so it would be good to keep coverage for them too (in fact, we might even want to have tests for them, if that makes sense).

Was it causing issues having them run through coverage?

andy-armstrong:
:+1: Thanks for getting this going, @AlasdairSwan and @benpatterson! My only question is whether we should leave coverage on the specs and spec_helpers.

AlasdairSwan:
@andy-armstrong I'm not even sure what it would mean for a test file to have test coverage? Wouldn't it be basically performing linting to make sure the code in the file is all used? If the helpers are specific to helping execute tests it seems a similar issue.

test/karma.ci.conf.js
AlasdairSwan: @andy-armstrong there were no issues with running them I just thought it was best practice to exclude test files from code coverage reports. If the helpers are part of the public API and used for other (non-test) things I'm fine with including them in the coverage report, but I'm still not convinced that Jasmine tests should be included.

andy-armstrong: I don't see the downside to including them, and it is occasionally useful to be able to see that certain tests and/or test files didn't get run. @benpatterson I defer to you on this. What is the policy we've used in other repos?

andy-armstrong:
@AlasdairSwan I asked in the TestEng chat room, and JZ said the following:

> yes, code coverage for tests is included in the platform metrics. The coverage config is here: https://github.com/edx/edx-platform/blob/master/.coveragerc

AlasdairSwan:
Thanks for following up on that @andy-armstrong, I will increase the coverage and merge this PR.

benpatterson:
:tada:

PR #46, opened 2016-03-26
Allow utilities to be loaded without RequireJS

andy-armstrong:
This commit adds the ability for utility classes to be loaded directly without RequireJS. If the class is loaded and RequireJS is not found, then the class will instead be installed into the edx namespace. For example, StringUtils can be accessed as ``edx.StringUtils``. 

The intention is that this will be useful for the LMS which is not fully RequireJS enabled. In particular, XBlocks and XModules have no ability to use RequireJS, so need global access to these utilities. This capability is being added to edx-platform in this PR: https://github.com/edx/edx-platform/pull/11904

In addition, HtmlUtils.template has been extended to pass StringUtils as a context parameter in addition to HtmlUtils itself.

@AlasdairSwan @cahrens please review.

FYI @bjacobel @dan-f @tobz @robrap @efischer19


edx-webhook:
## PR Cover Letter

## Sandbox

## Testing
- [ ] i18n
- [ ] RTL
- [ ] Analytics
- [ ] Performance

## Reviewers
- [ ] Code review #1
- [ ] Code review #2
- [ ] UX review
- [ ] Accessibility review
- [ ] Product review

## DevOps assistance
- configuration changes
- feature flags
- migrations
- [ ] File a support ticket

## Post-review
- [ ] Squash commits

andy-armstrong:
@AlasdairSwan I've been unable to debug why the Travis CI build keeps failing. Maybe we can look at this together on Monday,

src/js/utils/constants.js
```diff
@@ -1,19 +1,25 @@
...
+                left: 37,
+                up: 38,
+                right: 39,
+                down: 40
+            }
+        };
+    });
+}).call(
+    this,
+    (typeof define !== 'undefined' && define) ||
```
cahrens: You shouldn't need the RequireJS check-- I was imagining you would use the exact syntax expected by the optimizer.

We use the namespace property, and the optimizer will rename things accordingly (assuming it is the expected syntax).

http://requirejs.org/docs/faq-advanced.html#rename

src/js/utils/global-loader.js
```diff
@@ -0,0 +1,52 @@
...
+(function() {
+    'use strict';
+
+    window.edx = window.edx || {};
+
+    window.edx.GlobalLoader = (function() {
+        var registeredModules = {};
+
+        // Register standard libraries
+        registeredModules.jquery = $;
```
cahrens: Why is this necessary?

andy-armstrong: It's to make it so that HtmlUtils can load its dependencies:

https://github.com/edx/edx-ui-toolkit/blob/andya/minor-api-fixes/src/js/utils/html-utils.js#L9

src/js/utils/constants.js
andy-armstrong: Thanks for the details about the syntax expected by the optimizer. I'll switch to that way of looking for ``define``. I don't understand why we don't need to use the ``RequireJS`` check though. This code is in the UI Toolkit and can be loaded in many different ways, including by code included in LMS factories where ``define`` is not defined and ``RequireJS.define`` is available. 

Let's discuss this in the morning and pin down right the combination of checks to be using in all these files. We might also want to recommend that LMS code use the same convention, rather than the current:

    define || RequireJS.define

I tried this originally here in the UI Toolkit but it doesn't work because ``define`` throws an undefined exception. The only reason it works in the LMS is due to this ugly hack where window is given a ``define`` property of undefined, rather than not having the property at all:

https://github.com/edx/edx-platform/blob/master/common/static/js/RequireJS-namespace-undefine.js#L7

For anyone reading along, here's an example of the standard idiom in the LMS:

https://github.com/edx/edx-platform/blob/master/lms/djangoapps/teams/static/teams/js/views/teams.js#L32

andy-armstrong:
@cahrens In my testing it seems that the ``RequireJS`` namespace is still needed for the moment. We should continue the discussion as to how we might be able to remove it. I'm okay with including this in the UI Toolkit because a) this is an edX-specific repo, and b) it is written so that it won't fail if it isn't available.

I've also updated the non-utils to use the same mechanism to ensure that they will work in edx-platform, except I didn't add the logic to let them load outside of RequireJS. I really don't want to support the entire UI Toolkit outside RequireJS, so I think of the utilities as being exceptions.

src/js/dropdown-menu/dropdown-menu-view.js
```diff
@@ -38,242 +39,247 @@ define([
...
-    }
+            });
+
+            return DropdownMenuView;
+        }
+    );
+}).call(
+    this,
+    (typeof define === 'function' && define.amd && define) ||
+    (typeof RequireJS !== 'undefined' && RequireJS.define)
```
AlasdairSwan: @andy-armstrong the diffs are showing weirdness for these files, is the only really difference the first few and last few lines of the file?

andy-armstrong: Try viewing the diffs with ``?w=1`` to see the non-whitespace changes:

https://github.com/edx/edx-ui-toolkit/pull/46/files?w=1

AlasdairSwan: Thanks for the tip, that's a lot less noisy

andy-armstrong: I don't know why GitHub keeps this feature so hidden. It is incredibly useful.

AlasdairSwan:
Once the Travis failures are resolved :+1: Do you want to find some time tomorrow to debug (sorry today was a lot busier than I thought it would be)?

src/js/dropdown-menu/dropdown-menu-view.js
```diff
@@ -38,242 +39,247 @@ define([
...
-        return DropdownMenuView;
-    }
+            });
+
+            return DropdownMenuView;
+        }
+    );
+}).call(
+    this,
+    (typeof define === 'function' && define.amd && define) ||
```
cahrens: I find this difficult to read (the last && define). I would prefer if clauses.

dan-f: Also, isn't that last `&& define` redundant with `typeof define === 'function'`?

andy-armstrong: The last ``&& define`` is what actually returns the value to be returned. It is a confusing syntax that only makes sense to ex-Lisp developers like me. :-) @cahrens and I have come up with a less concise but easier to comprehend version.

andy-armstrong:
@cahrens @AlasdairSwan @dan-f I've addressed the review feedback so now all that remains is to a) get the Travis builds to pass, and b) to try it out on a sandbox.

cahrens:
:+1: if it works. :)

andy-armstrong:
@cahrens @AlasdairSwan strangely the Travis build just passed without me changing anything...

AlasdairSwan:
Gotta love flaky Travis tests...

PR #48, opened 2016-03-29
Fix the Travis test failure

andy-armstrong:
After a lot of research, I stumbled upon this Stack Overflow answer:

http://stackoverflow.com/questions/20733090/karma-error-there-is-no-timestamp-for/21153198#21153198

The problem seems to be some combination of the Karma configuration, the RequireJS configuration and PhantomJS. I've cleaned up the Karma configuration so that it loads files in an optimal order, and so that it doesn't accidentally include random other JS files from node_modules. In particular, it loads Underscore immediately rather than via RequireJS, because other libraries assume that it can be accessed from the global namespace. I've also fixed the RequireJS configuration so that it shims JQuery since it is pre-loaded by Jasmine-JQuery.

I've also experimented with removing the PhantomJS browser from the Travis tests. I think it isn't necessary since we're testing with both Chrome and Firefox. My build passed once without it, but I've added it back for the moment as I'm not convinced that that was necessary.

@AlasdairSwan @bjacobel please review.

FYI @dan-f @dsjen @benpatterson @jzoldak

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,6 @@ before_install:
 
 install:
     - npm install
-    - bower install
 
 script: gulp test --ci
 
--- a/README.md
+++ b/README.md
@@ -2,8 +2,10 @@
 
 A JavaScript toolkit for building edX user interfaces.
 
-[![GitHub version](https://badge.fury.io/gh/edx%2Fedx-ui-toolkit.svg)](https://badge.fury.io/gh/edx%2Fedx-ui-toolkit)  [![npm version](https://badge.fury.io/js/edx-ui-toolkit.svg)](https://badge.fury.io/js/edx-ui-toolkit)  [![Bower version](https://badge.fury.io/bo/edx-ui-toolkit.svg)](https://badge.fury.io/bo/edx-ui-toolkit)
-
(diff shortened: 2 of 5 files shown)


edx-webhook:
## PR Cover Letter

## Sandbox

## Testing
- [ ] i18n
- [ ] RTL
- [ ] Analytics
- [ ] Performance

## Reviewers
- [ ] Code review #1
- [ ] Code review #2
- [ ] UX review
- [ ] Accessibility review
- [ ] Product review

## DevOps assistance
- configuration changes
- feature flags
- migrations
- [ ] File a support ticket

## Post-review
- [ ] Squash commits

andy-armstrong:
I've got a second green build with PhantomJS added back. Let me know whether folks think it is better to remove it to reduce future risk, or whether we prefer the additional testing.

bjacobel:
More tests are better. If we get more unexplained build failures we'll know what to remove first but I would say since it's green for now to keep PhantomJS in.

bjacobel:
Changes LGTM :thumbsup:

benpatterson:
FWIW PhantomJS is typically the fastest and therefore will expose flaky conditions/tests sooner.

AlasdairSwan:
:ship: it :+1:

andy-armstrong:
I've added back PhantomJS and re-run the build three times with no failures. Time to merge!

PR #53, opened 2016-04-05
Update getActivefilterfields, add getFilterFieldValue

dan-f:
@andy-armstrong @dsjen could you review this small improvement? It unblocks some work for Learner Analytics.

### DIFF
--- a/src/js/pagination/paging-collection.js
+++ b/src/js/pagination/paging-collection.js
@@ -302,14 +302,30 @@
             },
 
             /**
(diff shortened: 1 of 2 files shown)


dsjen:
LGTM!  My only concern is that the data structure returned by `getActiveFilterFields` is changing.  Is it correct that Insights is the only application using this?

andy-armstrong:
:+1:

dan-f:
correct, @dsjen. Insights is the only client to this code.

PR #54, opened 2016-04-05
Fix issues with karma hanging on exit

bjacobel:
See issue [here](https://github.com/karma-runner/karma-coverage/issues/209#issuecomment-205852972). `karma-coverage` with reporters `text` and `text-summary` and our setup for some reason has an issue where it will hang for ~30 seconds prior to exiting. Since coverage reports are only marginally helpful in development, this PR removes the `text-summary` reporter. (It doesn't affect Coveralls.)

Also tweaks a few other things around the Karma setup:
- adds ability to run tests with `npm test`, which means you don't need to install Gulp globally
- removes a library that wasn't doing anything
- fixes linting issues

cc @andy-armstrong

### DIFF
--- a/gulp/tasks/jscs.js
+++ b/gulp/tasks/jscs.js
@@ -5,7 +5,7 @@
         gulp = require('gulp'),
         paths = {
             lint: [
-                'test/**/*.js',
+                'test/**/!(jquery.simulate)*.js',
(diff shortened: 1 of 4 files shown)


andy-armstrong:
:+1: Thanks for tracking this down as it had been bothering me too.

PR #58, opened 2016-04-19
Add forJQuery function

robrap:
Before:
```
$element.before(HtmlUtils.ensureHtml(htmlSnippet).toString())
```
After:
```
$element.before(HtmlUtils.forJQuery(htmlSnippet))
```


robrap:
@andy-armstrong and I decided we will stick with:
```
$element.before(HtmlUtils.ensureHtml(htmlSnippet).toString())
```

PR #59, opened 2016-04-21
Upgrade AjaxHelpers to match edx-platform

andy-armstrong:
## Description

I've updated the AjaxHelpers class to adopt all the changes that have been made in edx-platform. My goal is to be able to then use the UI Toolkit's version in platform, and thus be able to remove the duplicated copy.

The one challenge I had is that ``AjaxHelpers.requests`` had been converted into a wrapper function that simplified its usage. To avoid having to change every caller in platform, I renamed the UI Toolkit version to be ``AjaxHelpers.withFakeRequests``.

## Post-review

## Reviewers

- [x] @bjacobel 
- [ ] @dan-f 

### DIFF
--- a/src/js/utils/spec-helpers/ajax-helpers.js
+++ b/src/js/utils/spec-helpers/ajax-helpers.js
@@ -15,85 +15,165 @@
 define(['sinon', 'underscore', 'URI'], function(sinon, _, URI) {
     'use strict';
 
-    var fakeRequests, expectRequest, expectJsonRequest, expectPostRequest, expectRequestURL,
-        respondWithJson, respondWithError, respondWithTextError, respondWithNoContent;
+    var XHR_READY_STATES, fakeServer, createFakeRequests, withFakeRequests, fakeRequests, currentRequest,
+        expectRequest, expectNoRequests, expectJsonRequest, expectPostRequest, expectRequestURL, skipResetRequest,
+        respond, respondWithJson, respondWithError, respondWithTextError, respondWithNoContent;
 
     /**
-     * Keep track of all requests to a fake server, and call `spec`
-     * with a reference to the server. Allows tests to respond to
-     * individual requests.
+     * An enumeration of valid XHR ready states.
      */
-    fakeRequests = function (spec) {
-        return function () {
-            var requests = [],
-                xhr = sinon.useFakeXMLHttpRequest(),
-                args = Array.prototype.slice.call(arguments);
-            xhr.onCreate = function(request) {
-                requests.push(request);
-            };
-            spec.apply(null, args.concat([requests]));
-            xhr.restore();
+    XHR_READY_STATES = {
+        UNSENT: 0,
+        OPENED: 1,
+        LOADING: 3,
+        DONE: 4
+    };
+
+    /* These utility methods are used by Jasmine tests to create a mock server or
+     * get reference to mock requests. In either case, the cleanup (restore) is done with
+     * an after function.
+     *
+     * This pattern is being used instead of the more common beforeEach/afterEach pattern
+     * because we were seeing sporadic failures in the afterEach restore call. The cause of the
+     * errors were that one test suite was incorrectly being linked as the parent of an unrelated
+     * test suite (causing both suites' afterEach methods to be called). No solution for the root
+     * cause has been found, but initializing sinon and cleaning it up on a method-by-method
+     * basis seems to work. For more details, see STUD-1264.
+     */
+
+    /**
+     * Get a reference to the mocked server, and respond to all requests
+     * with the specified response.
+     *
+     * @param {string} response the fake response.
+     * @returns {*} The current request.
+     */
+    fakeServer = function(response) {
+        var server = sinon.fakeServer.create();
+        afterEach(function() {
+            if (server) {
+                server.restore();
+            }
+        });
+        server.respondWith(response);
+        return server;
+    };
+
+    createFakeRequests = function() {
+        var requests = [],
+            xhr = sinon.useFakeXMLHttpRequest();
+
+        requests.currentIndex = 0;
+        requests.restore = function() {
+            if (xhr && xhr.hasOwnProperty('restore')) {
+                xhr.restore();
+            }
         };
+        xhr.onCreate = function(request) {
+            requests.push(request);
+        };
+
+        return requests;
+    };
+
+    /**
+     * Keep track of all requests to a fake server, and return a reference to the Array.
+     * This allows tests to respond to individual requests.
+     *
+     * @returns {Array} An array tracking the fake requests.
+     */
+    fakeRequests = function() {
+        var requests = createFakeRequests();
+        afterEach(function() {
+            requests.restore();
+        });
+        return requests;
+    };
+
+    /**
+     * Wraps a test function so that it is invoked with an additional parameter
+     * containing an array of fake HTTP requests.
+     *
+     * @param {function} test A function to be invoked with the fake requests.
+     * @returns {function} A wrapped version of the input function.
+     */
+    withFakeRequests = function(test) {
+        return function() {
+            var requests = createFakeRequests(),
+                args = Array.prototype.slice.call(arguments);
+            test.apply(null, args.concat([requests]));
+            requests.restore();
+        };
+    };
+
+    /**
+     * Returns the request that has not yet been responded to. If no such request
+     * is available then the current test will fail.
+     *
+     * @param {object} requests an array of fired sinon requests
+     * @returns {*} The current request.
+     */
+    currentRequest = function(requests) {
+        expect(requests.length).toBeGreaterThan(requests.currentIndex);
+        return requests[requests.currentIndex];
     };
 
     /**
      * Expect that a request was made as expected.
      *
-     * @param requests an array of fired sinon requests
-     * @param method the expected method of the request
-     * @param url the expected url of the request
-     * @param body the expected request body
-     * @param requestIndex the expected index of the request in the
-     *     array (higher indices occur later).  Defaults to
-     *     requests.length - 1.
+     * @param {object} requests an array of fired sinon requests
+     * @param {string} method the expected method of the request
+     * @param {string} url the expected url of the request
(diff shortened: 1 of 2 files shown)


bjacobel:
:thumbsup:

src/js/utils/spec-helpers/ajax-helpers.js
```diff
@@ -1,9 +1,20 @@
 define(['sinon', 'underscore', 'URI'], function(sinon, _, URI) {
     'use strict';
 
-    var fakeRequests, expectRequest, expectJsonRequest, expectPostRequest, expectRequestURL,
+    var XML_HTTP_READY_STATES, fakeServer, createFakeRequests, withFakeRequests, fakeRequests, currentRequest,
```
dan-f: Nit: perhaps rename `XML_HTTP_READY_STATES` to `XHR_READY_STATES` because it's a bit shorter, and it masks XML as "X" :). I think the whole "XML" part of "XHR" can be misleading these days when XHRs are used for all kinds of content-types.

src/js/utils/spec-helpers/ajax-helpers.js
```diff
@@ -1,9 +1,20 @@
...
+        expectRequest, expectNoRequests, expectJsonRequest, expectPostRequest, expectRequestURL, skipResetRequest,
```
dan-f: It seems like theres's some redundancy between these variables and the exported module object at the bottom of the file. You could cut down on this long list of variable definitions and remove the need to keep it in sync with the module object by having two variables, `XHR_READY_STATES`, and `AjaxHelpers`. `AjaxHelpers` could then be the module object which is returned, and these functions could be defined as methods directly on that object.

src/js/utils/spec-helpers/ajax-helpers.js
```diff
@@ -17,204 +28,264 @@ define(['sinon', 'underscore', 'URI'], function(sinon, _, URI) {
...
+
+    /**
+     * Returns the request that has not yet been responded to. If no such request
+     * is available then the current test will fail.
+     *
+     * @param {object} requests an array of fired sinon requests
+     * @returns {*} The current request.
+     */
+    currentRequest = function(requests) {
+        expect(requests.length).toBeGreaterThan(requests.currentIndex);
```
dan-f: This makes the assumption that we're testing with jasmine, or a test framework which implements `expect`. I think we can simply throw an error, which would fail the test as well.

dan-f:
I see that we're using `currentRequest` in a lot of places. I like that it simplifies things, but there are real cases in which code will fire multiple requests concurrently. In those cases it would be useful for these APIs to support a way of specifying which request to operate on. This could be by passing in an optional request object or optional request index.

src/js/utils/spec-helpers/ajax-helpers.js
```diff
@@ -17,204 +28,264 @@ define(['sinon', 'underscore', 'URI'], function(sinon, _, URI) {
...
-        if (_.isUndefined(requestIndex)) {
-            requestIndex = requests.length - 1;
-        }
-        request = requests[requestIndex];
+     * @param {object} requests an array of fired sinon requests
+     * @param {string} method the expected method of the request
+     * @param {string} url the expected url of the request
+     * @param {object} jsonRequest the expected request body as an object
+     */
+    expectJsonRequest = function(requests, method, url, jsonRequest) {
```
dan-f: Nit: I think `bodyObject` or simply `body` might be clearer.

src/js/utils/spec-helpers/ajax-helpers.js
```diff
@@ -17,204 +28,264 @@ define(['sinon', 'underscore', 'URI'], function(sinon, _, URI) {
...
+     * @param {string} url the expected url of the request
+     * @param {string} body the expected body of the request
      */
-    expectPostRequest = function(requests, url, body, requestIndex) {
-        var request;
-        if (_.isUndefined(requestIndex)) {
-            requestIndex = requests.length - 1;
-        }
-        request = requests[requestIndex];
+    expectPostRequest = function(requests, url, body) {
```
dan-f: We should implement this on top of `expectRequest` to avoid duplicated logic:

```javascript
expectPostRequest = function(requests, url, body) {
    return expectRequest(requests, 'POST', url, body);
};
```

src/js/utils/spec-helpers/ajax-helpers.js
```diff
@@ -17,204 +28,264 @@ define(['sinon', 'underscore', 'URI'], function(sinon, _, URI) {
...
-     *     array (higher indices occur later).  Defaults to
-     *     requests.length - 1.
-     */
-    respondWithError = function(requests, statusCode, jsonResponse, requestIndex) {
-        if (_.isUndefined(requestIndex)) {
-            requestIndex = requests.length - 1;
-        }
+     * @param {object} jsonResponse an object to be serialized to the response
+     */
+    respondWithError = function(requests, statusCode, jsonResponse) {
```
dan-f: I like that `statusCode` and `jsonResponse` are optional, but this construct enforces that the consumer must supply `statusCode` if they're going to supply `jsonResponse`. How about moving both into an `options` hash?

dan-f:
I feel like there's a bit of asymmetry between `respondWithJson`, `respondWithError`, `respondWithTextError`, and `respondWithNoContent`.
- `respondWithJson` explicitly calls out that it converts an object to a JSON string in the response while `respondWithError` does this implicitly.
- There's a `respondWithTextError`, but no `respondWithText`.

Furthermore, I wonder if we could consolidate this logic into a single helper method, `respond`:

(This same pattern can be extended to `expect*Request`)

```javascript
var respond = function (requests, status, body) {
    var responseBody;

    if (_.isString(body)) {
        responseBody = body;
    } else if (_.isObject(body)) {
        responseBody = JSON.stringify(body);
    } else if (_.isUndefined(body)){
        responseBody = '';
    } else {
        // maybe throw an error...
    }

    // ...
};

// Respond with JSON:
AjaxHelpers.respond(requests, 200, {foo: 'bar'});

// Respond with text:
AjaxHelpers.respond(requests, 200, 'foo, bar');

// Respond with error, no body:
AjaxHelpers.respondWithError(requests, 500);

// Respond with error, JSON body:
AjaxHelpers.respondWithError(requests, 500, {error_code: 'Server error: something awful happened.'});

// Respond with error, text body:
AjaxHelpers.respondWithError(requests, 500, "We're sorry, but our servers are on fire!");
```

dan-f:
@andy-armstrong thanks for doing this! I finished my first pass on this PR.

I really like the idea of strengthening the UITK's `AjaxHelpers` module! Since this repository codifies our best practices and patterns, I'd like to make sure that these APIs are as flexible and consistent as possible before merging them in.

andy-armstrong:
Thanks for all the great feedback @dan-f. I'm just getting to your comments now. 

I love your suggested simplification of the various fake response functions. It will make it harder to just start using in edx-platform, so I think I might keep the other methods but mark them as deprecated (if that works for you). I don't want to go too long without removing edx-platform's copy of this class, because it will just keep diverging.

andy-armstrong:
Regarding ``currentRequest``, it was actually deliberate to force developers to handle each request in turn. You can see the original platform PR here:

https://github.com/edx/edx-platform/pull/9858

I'd rather not allow folks to pick and choose which request they handle. If you have three requests, you just need to verify and respond to each one in turn. Let me know what you think.

src/js/utils/spec-helpers/ajax-helpers.js
andy-armstrong: Done

src/js/utils/spec-helpers/ajax-helpers.js
andy-armstrong: This file has many calls to ``expect`` already, and it gives a cleaner failure message than throwing an exception. I'm going to leave it as is.

src/js/utils/spec-helpers/ajax-helpers.js
andy-armstrong: This isn't straightforward as the analysis of the body is very different in the two cases. I'm going to leave it as is for now.

andy-armstrong:
@bjacobel @cahrens I've finally got back to this PR. I've merged in the latest changes from edx-platform, and I've addressed Dan's feedback. Please review.

andy-armstrong:
Hmmm, I don't get any of these failures locally... I'm investigating.

andy-armstrong:
@bjacobel @cahrens OK, I accidentally re-enabled some tests that clearly were still failing. I've left those lines unchanged and now I have a green build again.

src/js/utils/spec-helpers/ajax-helpers.js
```diff
@@ -15,208 +15,302 @@
...
+        afterEach(function() {
+            if (server) {
+                server.restore();
+            }
+        });
+        server.respondWith(response);
+        return server;
+    };
+
+    /**
```
cahrens: I think this comment get left behind, or duplicated?

src/js/utils/spec-helpers/ajax-helpers.js
```diff
@@ -15,208 +15,302 @@
...
+        };
+        xhr.onCreate = function(request) {
+            requests.push(request);
+        };
+
+        return requests;
+    };
+
+    /**
+     * Keep track of all requests to a fake server, and return a reference to the Array.
```
cahrens: Here  is the same block of documentation for the third time.

src/js/utils/spec-helpers/ajax-helpers.js
```diff
@@ -15,208 +15,302 @@
...
+        return server;
+    };
+
+    /**
+     * Keep track of all requests to a fake server, and
+     * return a reference to the Array. This allows tests
+     * to respond for individual requests.
+     */
+
+    /**
```
cahrens: Here is the same block of documentation for the second time.

bjacobel:
LGTM.

cahrens:
👍 After you clean up the documentation issue I noticed.

PR #61, opened 2016-04-29
Implement searching on top of filtering

dan-f:
## Description

This PR introduces `getSearchString` and `hasActiveSearch` by re-implementing searching on top of filtering, thus combining unnecessarily divergent code paths for consistency and to prevent future regressions.

## Post-review

## Reviewers

- [x] @peter-fogg 
- [x] @bjacobel  

### DIFF
--- a/src/js/pagination/paging-collection.js
+++ b/src/js/pagination/paging-collection.js
@@ -32,7 +32,6 @@
 
             sortableFields: {},
 
-            searchString: '',
             filterableFields: {},
 
             state: {
@@ -46,7 +45,6 @@
                 pageSize: 'page_size',
                 sortKey: 'order_by',
                 order: 'sort_order',
-                text_search: function () { return this.searchString || null; }
             },
 
             constructor: function (models, options) {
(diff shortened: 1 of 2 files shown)


src/js/pagination/paging-collection.js
```diff
@@ -302,19 +300,23 @@
             },
 
             /**
-             * Gets an object of currently active (applied) filters.
+             * Gets an object of currently active (applied) filters excluding
+             * the active search.
              * @returns {Object} An object mapping the names of
```
bjacobel: Should update this jsdoc string to describe the new argument to this method.

bjacobel:
LGTM after the doc fix :thumbsup:

andy-armstrong:
👍 nice clean up! 

Would you be willing to write test to cover the last few lines from this file? It is so close to 100%!

https://coveralls.io/builds/5993304/source?filename=src%2Fjs%2Fpagination%2Fpaging-collection.js

src/js/pagination/paging-collection.js
```diff
@@ -343,11 +345,9 @@
                     value = value.join(',');
                 }
                 this.filterableFields[fieldName].value = value;
-                if (_.isUndefined(this.queryParams.fieldName)) {
```
peter-fogg: I'm not totally clear on why the conditional is being removed here -- does something about the refactor make it unnecessary? Was it ever needed?

peter-fogg:
👍

src/js/pagination/paging-collection.js
dan-f: It wasn't ever needed; I mistakenly added it in order to not have to redefine the same function over and over. However I ran into some edge cases in my consumer code where changing the value of a filter didn't actually result in the new (rather the old) value being sent in the subsequent request. Removing this conditional fixed that issue. Without diving too deep into the code or thinking about it too hard I'm pretty sure it had to do with JS's odd closure implementation.

src/js/pagination/paging-collection.js
dan-f: Good catch!

dan-f:
@bjacobel I clarified that JSDoc comment.
@andy-armstrong I *think* that this file should be 100% covered now, but I couldn't get the coverage report to update locally. I'll re-ping you when I confirm and am seeking the final 👍  :)

bjacobel:
LGTM :thumbsup:

dan-f:
@andy-armstrong - coverage is now at 100%!

andy-armstrong:
Awesome @dan-f. 👍

PR #66, opened 2016-05-06
Add a doc build to the default gulp task

andy-armstrong:
## Description

Add the doc build to the default gulp task, and change Travis to run the default task rather than just the tests. This is to ensure that all the gulp tasks are still working.

## Post-review

## Reviewers
- [x] @bjacobel 

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: node_js
 sudo: false
 
 node_js:
-    - "4.1"
+    - 4
 
 before_install:
     - export CHROME_BIN=chromium-browser
@@ -11,9 +11,11 @@ before_install:
     - sh -e /etc/init.d/xvfb start
 
 install:
+    - rvm install 2.2.2
     - npm install
+    - bundle install
 
-script: gulp test --ci
+script: gulp default --ci
 
 branches:
     only:
--- a/gulp/tasks/clean.js
+++ b/gulp/tasks/clean.js
@@ -1,0 +1,20 @@
+'use strict';
+
+var gulp    = require('gulp'),
+    config  = require('../config'),
+    del     = require('del');
+
+gulp.task('clean', function() {
+    return del([
+        // Remove the Jekyll site
+        config.documentation.targetDir,
+
+        // Remove the preview site
+        config.documentation.previewTargetDir,
+
+        // Remove the JSDoc generated markdown
+        config.documentation.testing.output,
+        config.documentation.utilities.output,
+        config.documentation.views.output
+    ]);
+});
--- a/gulp/tasks/default.js
+++ b/gulp/tasks/default.js
@@ -1,5 +1,10 @@
-(function() {
-    'use strict';
-    var gulp = require('gulp');
-    gulp.task('default', ['test', 'lint', 'jscs', 'coverage']);
-}());
+'use strict';
+
+var gulp = require('gulp');
+
+gulp.task('default', [
+    'lint',
+    'jscs',
+    'test',
+    'doc-build'
+]);
--- a/gulp/config.js
+++ b/gulp/config.js
@@ -1,5 +1,7 @@
 module.exports = {
     documentation: {
+        targetDir: './_site',
+        previewTargetDir: './_preview_site',
         testing: {
             sources: [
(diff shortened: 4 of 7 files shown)


andy-armstrong:
@bjacobel @dan-f @dsjen @AlasdairSwan - can someone give me a quick review of this. Thanks!

.travis.yml
```diff
@@ -6,14 +6,16 @@ node_js:
     - "4.1"
 
 before_install:
+    - rvm install 2.2.2
```
dan-f: Why put this in `before_install` rather than `install`?

gulp/tasks/clean.js
```diff
@@ -0,0 +1,20 @@
...
+        config.documentation.targetDir,
+
+        // Remove the preview site
+        config.documentation.previewTargetDir,
+
+        // Remove the JSDoc generated markdown
+        config.documentation.testing.output,
+        config.documentation.utilities.output,
+        config.documentation.views.output
+    ]);
```
dan-f: Perhaps also clean up coverage files?

gulp/tasks/default.js
```diff
@@ -1,5 +1,10 @@
...
-    'use strict';
-    var gulp = require('gulp');
-    gulp.task('default', ['test', 'lint', 'jscs', 'coverage']);
-}());
+'use strict';
+
+var gulp = require('gulp');
+
+gulp.task('default', [
+    'doc-build',
```
dan-f: Are these run in order? If so, does a single task failure stop the whole list of tasks? If so, I think test/lint/jscs should be above doc-build, since it seems pointless to build docs if the test suite is going to fail.

gulp/tasks/default.js
```diff
@@ -1,5 +1,10 @@
...
+    'lint',
+    'jscs',
```
dan-f: I know this isn't related to the current PR, but I've seen this pattern now in a few of our repos (including Insights) where we separate "lint" from "jscs". I've always felt like they can be bundled together, e.g. "lint" calls `jshint` as well as `jscs`. If they're going to remain separate though, I think it's worth renaming `jscs` to `style` so that there's symmetry between `lint`/`style` - currently one of the task names represents a task to perform while the other is just the name of a tool.

.travis.yml
dan-f: Alternatively, it might be more travis-like to do something like:

```yaml
languages:
  - node_js
  - ruby

node_js:
  - 4.1

rvm:
  - 2.2.2
```

andy-armstrong: @dan-f Does that work? Everything I read said that you had to pick a single language. For example, here's where I took my configuration from:

http://stackoverflow.com/questions/31235146/how-to-run-node-js-and-ruby-tests-within-one-project-on-travis-ci

I'll try your suggestion which will be much cleaner if it works.

dan-f: It does work! Though the key is still "language", not "languages" - my mistake.

https://github.com/edx/edx-analytics-dashboard/blob/master/.travis.yml#L1-L5

andy-armstrong: It seems that Travis does not support multiple languages yet:

https://github.com/travis-ci/travis-ci/issues/4090

dan-f: If it's not yet implemented, I wonder why it's working for us in Insights. @benpatterson added that config - Ben, thoughts?

andy-armstrong: @dan-f I tried specifying two languages in my other PR but it doesn't work. I think that Travis must just pick the first one, and that happens to work for you.

gulp/tasks/default.js
andy-armstrong: I believe they are asynchronous, but I've put it at the end anyway.

gulp/tasks/default.js
andy-armstrong: Note that we're going to switch over to ESLint which will then remove the need for JSCS entirely:

https://openedx.atlassian.net/browse/FEDX-164

gulp/tasks/clean.js
andy-armstrong: Do you happen to know where they go?

bjacobel:
:shipit:

PR #67, opened 2016-05-10
update parseState to update all state properties

ehteshamkafeel:
## Description
These changes add the ability to set the state from response without fetching the collection from server. Before these changes collection state would only update after the response has come from server. So if we set the response during instantiating or in `set` and `reset` methods of collection, the state wouldn't update other than the `totalRecords` and `totalPages`.

These changes has overridden two more query parameters value `totalRecords` and `totalPages` and removed the overridden `parseState` method in favor of `PageableCollection`'s more complete version of the method that ensures that all passed response gets evaluated and state is updated accordingly. For example if now we can set the current page during instantiation like following.

```javascript
var collection = new PagingCollection({
    count: 10,
    page: 2,
    page_size: 3,
   num_pages: 4,
    results: []
}, {parse: true}); 
```

These changes also include some getters to get state. I thought it would be better to not touch the state directly, ideally state shouldn't be changed directly.

## Testing Checklist

## Post-review

## Reviewers
- [x] @dan-f
- [x] @andy-armstrong 

### DIFF
--- a/src/js/pagination/paging-collection.js
+++ b/src/js/pagination/paging-collection.js
@@ -46,8 +46,10 @@
             queryParams: {
                 currentPage: 'page',
                 pageSize: 'page_size',
+                totalRecords: 'count',
+                totalPages: 'num_pages',
                 sortKey: 'order_by',
-                order: 'sort_order',
+                order: 'sort_order'
             },
 
             constructor: function (models, options) {
@@ -77,15 +79,6 @@
                 return PageableCollection.prototype.parse.call(this, modifiedResponse, options);
             },
 
-            /* jshint unused:false */
-            /**
-             * Parses state from the server response.  Used only by
-             * backbone.paginator.
-             */
-            parseState: function (response, queryParams, state, options) {
-                return {totalRecords: response[0].count, totalPages: response[0].num_pages};
-            },
-
             /**
              * Returns the current page number as if numbering starts on
              * page one, regardless of the indexing of the underlying
@@ -95,6 +88,34 @@
              */
             getPageNumber: function () {
                 return this.state.currentPage + (1 - this.state.firstPage);
+            },
+
+            /**
+             * Returns the total pages of the collection based on
+             * total records and page size
+             *
+             * @returns {integer} Total number of pages.
+             */
+            getTotalPages: function () {
+                return this.state.totalPages;
+            },
+
+            /**
+             * Returns the total number of records the collection has
+             *
+             * @returns {integer} Total number of records.
+             */
+            getTotalRecords: function () {
+                return this.state.totalRecords;
+            },
+
+            /**
+             * Returns the number of records per page
+             *
+             * @returns {integer} Number of records per page.
+             */
+            getPageSize: function () {
+                return this.state.pageSize;
             },
 
             /**
--- a/src/js/pagination/specs/paging-collection-spec.js
+++ b/src/js/pagination/specs/paging-collection-spec.js
@@ -65,6 +65,53 @@ define(['jquery',
                 expect(newCollection.queryParams.currentPage).toBe('page');
             });
 
+            SpecHelpers.withData({
+                'correctly sets state when parse is set to true after instantiating': [null, null],
+                'correctly sets state when parse is set to true in set method': ['set', {
+                    page: 3,
+                    numPages: 5,
+                    pageSize: 10
+                }],
+                'correctly sets state when parse is set to true in reset method': ['reset', {
+                    page: 3,
+                    numPages: 3,
+                    pageSize: 15
+                }]
+            }, function (method, expectedState) {
+                var newCollection;
+                newCollection = new PagingCollection({
+                    count: 43,
+                    page: 2,
+                    num_pages: 6,
+                    page_size: 8,
+                    results: []
+                }, {parse: true});
+
+                expect(newCollection.getPageSize()).toBe(8);
+                expect(newCollection.getPageNumber()).toBe(2);
+                expect(newCollection.getTotalRecords()).toBe(43);
+                expect(newCollection.getTotalPages()).toBe(6);
+
+                if (method) {
+                    newCollection[method]({
+                        count: 43,
+                        page: expectedState.page,
+                        num_pages: expectedState.numPages,
+                        page_size: expectedState.pageSize,
+                        sort_order: -1,
+                        order_by: 'testcol',
+                        results: []
+                    }, {parse: true});
+
+                    expect(newCollection.getPageSize()).toBe(expectedState.pageSize);
+                    expect(newCollection.getPageNumber()).toBe(expectedState.page);
+                    expect(newCollection.getTotalPages()).toBe(expectedState.numPages);
+                    expect(newCollection.getTotalRecords()).toBe(43);
+                    expect(newCollection.state.sortKey).toBe('testcol');
+                    expect(newCollection.state.order).toBe(-1);
+                }
+            });
+
             it('can register sortable fields', function () {
                 collection.registerSortableField('test_field', 'Test Field');
                 expect('test_field' in collection.sortableFields).toBe(true);


src/js/pagination/paging-collection.js
```diff
@@ -83,7 +83,15 @@
...
+                state.totalRecords = response[0].count;
+                state.totalPages = response[0].num_pages;
+
+                // Remove already processed response
+                delete response.count;
+                delete response.num_pages;
+
+                // Update the state with response
+                return PageableCollection.prototype.parseState.call(this, response, queryParams, state, options);
             },
```
ehteshamkafeel: While doing a reset or set on collection directly the state doesn't set with server returned or passed data. So calling parent's (PageableCollection) parseState ensures that state is in sync.

src/js/pagination/paging-collection.js
```diff
@@ -98,6 +106,35 @@
...
 
             /**
+             * Returns the totalPages of the collection based on
```
dan-f: Could you please change `totalPages` here to `total pages`? The idea is that whoever is reading this shouldn't have to know about the internals of how the `PagingCollection` works.

src/js/pagination/paging-collection.js
```diff
@@ -98,6 +106,35 @@
...
+             *
+             * @returns {integer} The current page number.
+             */
+            getTotalPages: function () {
+                return this.state.totalPages;
+            },
+
+            /**
+             * Returns the total number of records the collection has
+             * on server
```
dan-f: Would you mind removing "on server"? It is assumed that the `PagingCollection` speaks to a server.

src/js/pagination/paging-collection.js
```diff
@@ -98,6 +106,35 @@
             },
 
             /**
+             * Returns the totalPages of the collection based on
+             * totalRecords and PageSize
+             *
+             * @returns {integer} The current page number.
```
dan-f: This parameter doc is incorrect.

src/js/pagination/paging-collection.js
```diff
@@ -98,6 +106,35 @@
...
+             */
+            getTotalPages: function () {
+                return this.state.totalPages;
+            },
+
+            /**
+             * Returns the total number of records the collection has
+             * on server
+             *
+             * @returns {integer} The current page number.
```
dan-f: This parameter doc is also incorrect.

src/js/pagination/paging-collection.js
```diff
@@ -98,6 +106,35 @@
...
+             */
+            getTotalRecords: function () {
+                return this.state.totalRecords;
+            },
+
+            /**
+             * Returns the number of records per page
+             *
+             * @returns {integer} The current page number.
```
dan-f: This parameter doc is also incorrect.

src/js/pagination/paging-collection.js
dan-f: Could you add a test case to prove this and prevent further regressions?

dan-f:
Thanks for the contribution, @ehteshamkafeel! 

I'd like to ask you to do a few things before I re-review:
- Add some tests for your changes. You can read more about our testing requirements in the [open edX contributing docs](https://edx.readthedocs.io/projects/edx-developer-guide/en/latest/process/index.html).
- Please update the description of this PR to describe the changes you've made. Currently it appears to be the default PR template that we use. You can delete any non-relevant sections.

Thanks!

ehteshamkafeel:
@dan-f yeah, I am actually writing tests. This PR isn't yet ready for a review yet that's why I didn't tag anyone. Thanks for your reviews though, makes perfect sense.

dan-f:
I see. Sorry to have interrupted your workflow! I thought I was tagged by you, but it looks like it was just @mention-bot doing its job :)

andy-armstrong:
👍  Looks good to me. Here are the steps to release the UI Toolkit:

https://openedx.atlassian.net/wiki/display/FEDX/UI+Toolkit+release+process

src/js/pagination/paging-collection.js
```diff
@@ -83,7 +83,15 @@
              * backbone.paginator.
              */
             parseState: function (response, queryParams, state, options) {
-                return {totalRecords: response[0].count, totalPages: response[0].num_pages};
+                state.totalRecords = response[0].count;
```
dan-f: This shouldn't mutate `state`. See: https://backbone-paginator.github.io/backbone.paginator/api/#parseState - "This method MUST return a new state object instead of directly modifying the PageableCollection#state object. The behavior of directly modifying PageableCollection#state is undefined."

Please follow the previous pattern; construct and return a new object.

ehteshamkafeel: "This method MUST return a new state object instead of directly modifying the PageableCollection#state object. The behavior of directly modifying PageableCollection#state is undefined."

@dan-f Well what this statement tries to convey is that we must return a state object instead of just  modifying the state and returning nothing in which case state is undefined. So this state that is being passed to this function is being passed a copy not the actual state object. So it's safe to modify and return the state. This is not the collection maintained state.

ehteshamkafeel:
@andy-armstrong @dan-f I have simplified my changes by removing the overridden `parseState` method in favor of more complete version of `PageableCollection`'s parseState. I have overrridden two more query parameters values in the state object's declaration in `PagingCollection`.

src/js/pagination/paging-collection.js
```diff
@@ -83,7 +83,16 @@
...
+                newState.totalRecords = response[0].count;
+                newState.totalPages = response[0].num_pages;
+
+                // Remove already processed response
+                delete response.count;
+                delete response.num_pages;
+
+                // Update the state with response
+                return PageableCollection.prototype.parseState.call(this, response, queryParams, newState, options);
             },
```
ehteshamkafeel: "This method MUST return a new state object instead of directly modifying the PageableCollection#state object. The behavior of directly modifying PageableCollection#state is undefined."

@dan-f Well what this statement tries to convey is that we must return a state object instead of just  modifying the state and returning nothing in which case state is undefined. So this state that is being passed to this function is being passed a copy not the actual state object. So it's safe to modify and return the state, parent is already doing a clone of the state. To be extra safe I have added a new statement to clone the passed state.

ehteshamkafeel: @dan-f can you please review this.

dan-f: Hi @ehteshamkafeel - Sorry for the lack of communication. I should have updated you the other day. I've been busy with development and code reviews for the Analytics team. I'm hoping to re-review this afternoon or on Monday.

dan-f:
Hi @ehteshamkafeel - thanks for your patience on my review.

While I really appreciate your contributing to the `PagingCollection`, I'm feeling that this change is not necessary; the requested behaviors are already implemented.

`Backbone.paginator` already has an API to set `state` through its [constructor](https://backbone-paginator.github.io/backbone.paginator/api/#constructor):

```js
newCollection = new PagingCollection({
  results: [ ... ]
}, {
  state: {
    totalRecords: 10,
    currentPage: 2,
    totalPages: 3,
    pageSize: 4,
  },
  parse: true
});
```

Using this API will get the first part of your test to pass.

As for the second part, you call `Backbone.collection.set()` and `Backbone.collection.reset()` in order to set `state` values.  I would not expect this to work, since `Backbone.paginator.PageableCollection` does not override `set`. Furthermore, I'm not sure if we'd need such an API. Once instantiated, a collection should not mutate its own state. Since the collection represents a server's paginated API, the state should be parsed from server responses, not on-the-fly set manually on the client.

So, while I don't agree with the `parseState` changes (:-1:), the new getters get a :+1: from me!

ehteshamkafeel:
@dan-f Have you seen the original implementation of `parseState` in backbone.paginator's `PageableCollection`. I don't know why we have overridden this method. If we don't override the method we get the expected behavior and we don't need these changes. **So original `parseState` method does change state in `set` or `reset` if parse is set true.** We only needed to customise the `totalRecords` and `totalPages` keys is queryParams to make it work.

dan-f:
@ehteshamkafeel I cannot recall why `parseState` was overridden, but I think it happened when I migrated this module from backbone paginator < 1.0 to ~2.0. If we can get away with removing it, great.

ehteshamkafeel:
@dan-f I have removed the `parseState` method from `PagingCollection` and customised the queryParams for `totalPages` and `totalRecords`. I didn't remove the test and it's still passing. Please review.

src/js/pagination/paging-collection.js
```diff
@@ -46,8 +46,10 @@
             queryParams: {
                 currentPage: 'page',
                 pageSize: 'page_size',
+                totalRecords: 'count',
+                totalPages: 'num_pages',
```
dan-f: @ehteshamkafeel in https://github.com/edx/edx-ui-toolkit/pull/69/ you set `totalRecords ` and `totalPages ` to `null` to prevent them being sent to the server during `GET` requests. So these changes would be incompatible, right?

Why not just set `state` through the documented [`backbone.paginator.PageableCollection.constructor`](https://backbone-paginator.github.io/backbone.paginator/api/#constructor) API?

dan-f:
@ehteshamkafeel could you outline your particular use case?

src/js/pagination/paging-collection.js
ehteshamkafeel: @dan-f yeah, they would be incompatible. But I know how to fix it, I will do my changes in that and tag you for review.

ehteshamkafeel:
@dan-f yeah I can set state with constructor, but really now I don't know of any reason why this has to be like it is and what's wrong with this change.

Anyways I will discard these changes.

dan-f:
Nothing wrong with the changes, and please don't discard them! I'm just hoping to better understand the use case. I'm 👍  with the current changes, btw.

ehteshamkafeel:
@dan-f Thanks for the thumbs up. Actually if we keep it as is then I will have to do a lot of refactoring in the edx-platoform. Thanks again!

ehteshamkafeel:
@andy-armstrong you good with these changes?

andy-armstrong:
Yep, looks great to me. 👍

PR #69, opened 2016-05-20
[PagingCollection] Do not send unnecessary query parameters in request in

ehteshamkafeel:
## Description
These changes remove extra query parameters, defined in queryParams object's totalRecords and totalPages keys. Before these changes there were extra params going in the request like `total_entries=10&total_pages=5`. These params were being sent only when state object's `totalRecords` and `totalPages` were set. So on first request these params wouldn't go but they would on subsequent requests when state is set by the server response. These params have no meaning for server and are extraneous.

## Testing Checklist
- [ ] Write unit tests for all new features.

## Post-review

## Reviewers
- [ ] @dan-f 
- [ ] @andy-armstrong 

### DIFF
--- a/src/js/pagination/paging-collection.js
+++ b/src/js/pagination/paging-collection.js
@@ -48,6 +48,8 @@
                 pageSize: 'page_size',
                 sortKey: 'order_by',
                 order: 'sort_order',
+                totalPages: null,
+                totalRecords: null
             },
 
             constructor: function (models, options) {
--- a/src/js/pagination/specs/paging-collection-spec.js
+++ b/src/js/pagination/specs/paging-collection-spec.js
@@ -45,6 +45,13 @@ define(['jquery',
                 });
             };
 
+            var assertNotInQueryParams = function (requests, params) {
+                var urlParams = getUrlParams(requests[requests.length - 1]);
+                params.forEach(function (param) {
+                    expect(urlParams[param]).toBe(undefined);
+                });
+            };
+
             beforeEach(function () {
                 collection = new PagingCollection([], {state: {pageSize: 10}});
                 collection.url = '/test';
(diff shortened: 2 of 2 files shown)


src/js/pagination/paging-collection.js
```diff
@@ -48,6 +48,8 @@
                 pageSize: 'page_size',
                 sortKey: 'order_by',
                 order: 'sort_order',
+                totalPages: null,
```
andy-armstrong: I'm confused by this. I thought we don't want these two as query parameters? Also, what's the difference between totalRecords and total_entries?

src/js/pagination/specs/paging-collection-spec.js
```diff
@@ -89,6 +96,28 @@ define(['jquery',
                 }
             ));
 
+            it('doesn\'t send unnecessary parameters in request (total pages, total records)', AjaxHelpers.requests(
```
andy-armstrong: Why is the actual query parameter ``total_entries`` but the title says ``total records``?

src/js/pagination/specs/paging-collection-spec.js
```diff
@@ -89,6 +96,28 @@ define(['jquery',
...
+                        state: {
+                            pageSize: 10,
+                            totalPages: 10,
+                            totalRecords: 100
+                        }
+                    });
+
+                    newCollection.url = '/test';
+                    newCollection.fetch();
+                    assertNotInQueryParams(requests, ['total_entries', 'total_pages']);
```
andy-armstrong: I'd prefer that this be a positive test. Can you just use ``AjaxHelpers.expectRequest`` and enumerate exactly the parameters that are expected. This would then verify that these two particular parameters aren't specified, but also that strange future ones don't get added too.

src/js/pagination/paging-collection.js
ehteshamkafeel: yeah exactly, we don't want this. But they are included in the parent's `queryParams`. So every thing that is truthy in query params gets included in the query params. By specifically marking them null ensures that they won't include in the final query params. 
`totalRecords` and `totalPages` are the `state` attributes and they the value of these depends on what we have in queryParams. So e.g. if we have the following queryParams the value of `totalRecords` and `totalPages` will depend on the response's `count` and  `total_pages `key respectovely.

```javascript
queryParams: {
    totalRecords: 'count',
    totalPages: 'total_pages'
}
```

PageableCollection assumes that the `totalRecords = response's 'total_entries' and totalPages = 'total_pages' `. And we can override this `queryParams` object.

src/js/pagination/specs/paging-collection-spec.js
ehteshamkafeel: because I have tried to use the english terms not the actual things what we have in code. And because `total_entries` could be other string. e.g. we could have `all_records` instead of `total_entries`.

src/js/pagination/specs/paging-collection-spec.js
ehteshamkafeel: Good suggestion. Thanks.

src/js/pagination/paging-collection.js
dan-f: Yeah, for some reason backbone paginator believes that the client should be telling the server `totalRecords` and `totalPages` - https://backbone-paginator.github.io/backbone.paginator/api/#queryParams

I agree that there's no reason for the client to request `totalRecords` or `totalPages`.

dan-f:
@ehteshamkafeel I agree with @andy-armstrong's request for a positive test. Other than that, 👍  from me! Thank you :)

ehteshamkafeel:
@dan-f @andy-armstrong My [PR against backbone.paginator](https://github.com/backbone-paginator/backbone.paginator/pull/327) has been merged. So next release will include these changes, at that time we should update `backbone.paginator`.

dan-f:
great!

Issue #32, opened 2016-03-11
Proposal: Style Guide


dan-f:
I'd like to propose that we design an explicit style guide for this project.

## Motivation:
- Consistency makes new code approachable and easy to read
- Can help prevent common JS bugs
- Makes contributing more accessible; e.g. sane indentation means you can contribute no matter what editor you use
- Avoids ambiguity

## Scope
- Naming (file names, require module names, "class" names, variable names)
- AMD module declaration style
- Testing style
- Variable declarations (hoisting, functions as variables vs function statements)
- Indentation (number of spaces, how to break lines)
- Paren and brace style

## Examples

- Airbnb has a [comprehensive style guide](https://github.com/airbnb/javascript)
- Google [has one too](https://google.github.io/styleguide/javascriptguide.xml)

andy-armstrong:
Closing this in favor of https://openedx.atlassian.net/browse/DOC-2755

dan-f:
I don't agree with closing this ticket. That DOC ticket is clearly "high-level guidance" while a style guide is necessarily lower level syntax reference.

pdesjardins:
@dan-f You're right about the way that DOC ticket is labeled. However, as I worked on the developer's guide section for that documentation project, I started migrating the JavaScript style guide information into the dev guide (alongside our Python and Objective C style guides).

I think the JavaScript style information you're describing would fit with the draft JavaScript style guide content I'm trying to merge in https://github.com/edx/edx-documentation/pull/906.

You can see the draft here: http://draft-fedx-dev-doc.readthedocs.io/en/latest/style_guides/javascript_guidelines.html.

andy-armstrong:
@dan-f Sorry that I didn't give enough context when closing this ticket. The FedX team and @pdesjardins have had a lot of discussions over the past few months about the right way to consolidate all the JavaScript documentation that we have in multiple places. We settled on using the Developer Guide as the central repository, and so given that IMO we shouldn't be creating any new per-repository material unless it really is unique. I completely agree about the need for a centralized style guide a la Google and AirBnB, and this is why @pdesjardins is spearheading this work in the Developer Guide.

PR #71, opened 2016-06-01
Add a style/conventions guide for writing JS

dan-f:
## Description

I wanted to take an initial attempt at addressing https://github.com/edx/edx-ui-toolkit/issues/32. I understand that this guide may eventually live in a wiki somewhere, but it seemed appropriate to develop it here for now.

My goal is that once people agree on the guide we can then worry about implementation details of how to "deploy" it.

## Reviewers
- [ ] @dsjen
- [ ] @bjacobel 
- [ ] @andy-armstrong
- [ ] @AlasdairSwan 
- [ ] @pdesjardins 

... and everyone else. I really want this to have the 👍  from an audience much broader than just FEDX.

### DIFF
--- a/style-guide.md
+++ b/style-guide.md
@@ -1,0 +1,165 @@
+# edX JavaScript Conventions
+
+# TODO
+
+I'm looking for widespread agreement on everything contained in this document!
+In particular, I'd love feedback on the [Goals](#goals), [Style](#style), and
+[Conventions](#conventions) sections.
+
+Here are a few things that would be awesome to have:
+- [ ] define and link to a jshint/jscs/eslint config
+- [ ] clarify path towards ES2015
+
+## Motivation
+
+Code conventions are important.  It helps us intuitively approach new code,
+revisit old code, and share code with others.
+
+## Goals
+
+We hope to write JavaScript which is:
+
+1. Legible
+2. Friendly
+3. Predictable
+4. Semantic
+
+## About
+
+This guide is split up into two sections - [Style](#style) and
+[Conventions](#conventions).  The style section addresses aesthetic topics,
+while the conventions section addresses topics that have more semantic
+significance.
+
+## Note: ECMAScript Version
+
+Currently, edX only supports ES5.  In order to implement modules, we use AMD by
+way of [RequireJS](http://requirejs.org/).  We are moving towards ES2015
+however, and this guide will be updated when we get there.
+
+## Style
+### Indentation
+
+All JavaScript should be indented with 4 spaces.  edX uses this standard
+throughout our Python, HTML, SASS/CSS, and JavaScript code.
+
+e.g.
+
+```javascript
+function sayHello (name) {
+    return 'Hello, ' + name;
+}
+```
+
+### Semicolons
+
+Use them!
+
+### Line Length
+
+Break lines at `120` columns.
+
+### Strings
+
+Use single quotes for string literals, unless the string itself contains single
+quotes.
+
+e.g.
+
+```javascript
+var edX = 'edX',
+    thisGuide = "edX's JavaScript Style Guide";
+```
+
+### Comments
+
+If code does not read clearly on its own, consider breaking it up into smaller
+pieces. If things are still not clear, add a descriptive comment in the form of
+full-sentence(s).
+
+### Variables
+
+We like to write our code to reflect the fact that JavaScript hoists `var`
+declarations to the top-level environment of the current function. Therefore
+declare all variables for a particular function at the start of the function.
+
+e.g.
+
+```javascript
+var enrollUserInCourse = function (username, courseId) {
+    var userModel,
+        courseModel;
+
+    if (username) {
+        userModel = new UserModel({username: username});
+    } else {
+        throw new Error('No user specified');
+    }
+
+    if (courseId) {
+        courseModel = new CourseModel({courseId: courseId});
+    } else {
+        throw new Error('No course specified');
+    }
+
+    courseModel.enroll(userModel);
+}
+```
+
+### Brackets, Braces, and Parentheses
+
+"Cuddle" the second keyword of `if/else`, `do/while`, and `try/catch` statements
+onto the previous closing brace.
+
+e.g.
+
+```javascript
+if (/* condition */) {
+    // ...
+} else { // <- cuddle here!
+    // ...
+}
+
+do {
+    // ...
+} while (/* ... */); // <- cuddle here!
+
+try {
+    // ...
+} catch (error) { // <- cuddle here!
+    // ...
+}
+```
+
+Leave spaces around parenthesized expressions in control structures and function
+declarations.
+
+e.g.
+
+```javascript
+function isNaivelyEqual (arg1, arg2) {
+    if (arg1 === arg2) {
+        return true;
+    }
+    return false;
+}
+```
+
+## Conventions
+### Documentation
+
+Document all modules, constructors, functions, and methods with
+[JSDoc](http://usejsdoc.org/) comments.
+
+### Modules
+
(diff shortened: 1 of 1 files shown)


style-guide.md
```diff
@@ -0,0 +1,165 @@
+# edX JavaScript Conventions
+
+# TODO
```
dan-f: Note to eventually delete this section.

style-guide.md
```diff
@@ -0,0 +1,165 @@
...
+revisit old code, and share code with others.
+
+## Goals
+
+We hope to write JavaScript which is:
+
+1. Legible
+2. Friendly
+3. Predictable
+4. Semantic
```
dan-f: I used some buzzwords here :) hopefully we can come up with something less vague.

style-guide.md
```diff
@@ -0,0 +1,165 @@
...
+We like to write our code to reflect the fact that JavaScript hoists `var`
+declarations to the top-level environment of the current function. Therefore
+declare all variables for a particular function at the start of the function.
+
+e.g.
+
+```javascript
+var enrollUserInCourse = function (username, courseId) {
+    var userModel,
+        courseModel;
```
bjacobel: I'm against multiple variable instantiations in the same `var` statement, but I don't really have a good reason for it.

andy-armstrong:
@dan-f Sorry that we didn't tell you about this PR, but I figured you were leaving. :-(

https://github.com/edx/edx-documentation/pull/906

You can see the initial cut at JavaScript guidelines here:

http://draft-fedx-dev-doc.readthedocs.io/en/latest/style_guides/javascript_guidelines.html

It probably makes the most sense to create a new PR against edx-documentation once the initial PR lands, so that we can discuss the conventions.

👎  for this change as is because this shouldn't live in the UI Toolkit, so let's get it into the documentation.

FYI @pdesjardins

style-guide.md
```diff
@@ -0,0 +1,165 @@
...
+## About
+
+This guide is split up into two sections - [Style](#style) and
+[Conventions](#conventions).  The style section addresses aesthetic topics,
+while the conventions section addresses topics that have more semantic
+significance.
+
+## Note: ECMAScript Version
+
+Currently, edX only supports ES5.  In order to implement modules, we use AMD by
```
bjacobel: The discussion of the kind of modules we support is somewhat separate from our Javascript syntax support (ES5/ES2015) - we will probably support writing ES2015 soon, but will not change the way we do modules to CommonJS or ES6 modules for quite some time.

style-guide.md
```diff
@@ -0,0 +1,165 @@
+# edX JavaScript Conventions
+
+# TODO
+
+I'm looking for widespread agreement on everything contained in this document!
+In particular, I'd love feedback on the [Goals](#goals), [Style](#style), and
+[Conventions](#conventions) sections.
+
+Here are a few things that would be awesome to have:
+- [ ] define and link to a jshint/jscs/eslint config
```
bjacobel: :thumbsup: - Andy and I are thinking about taking this on for the hackathon. Most of the recommendations in this styleguide are compatible with the Airbnb ESLint config, which I like a lot.

style-guide.md
```diff
@@ -0,0 +1,165 @@
...
+- [ ] clarify path towards ES2015
```
bjacobel: :thumbsup: - this ties into our browser support SLA, which I would like to see us update this summer so that we can start writing some ES2015 without any Babel at all

style-guide.md
```diff
@@ -0,0 +1,165 @@
...
+}
+```
+
+### Semicolons
+
+Use them!
+
+### Line Length
+
+Break lines at `120` columns.
```
bjacobel: I believe - though this may no longer be true - that Github's "unified" diff view caps out at 118 characters, so I've used a line length of 117 at previous jobs for this reason

dan-f: Good call!

style-guide.md
```diff
@@ -0,0 +1,165 @@
...
+```javascript
+var edX = 'edX',
+    thisGuide = "edX's JavaScript Style Guide";
+```
+
+### Comments
+
+If code does not read clearly on its own, consider breaking it up into smaller
+pieces. If things are still not clear, add a descriptive comment in the form of
+full-sentence(s).
```
bjacobel: Full sentences is an interesting requirement, I think right now very few of my comments adhere to that rule. What's your thinking behind this?

dan-f:
@andy-armstrong thanks for the feedback. I mentioned at the top level that I agree that this shouldn't necessarily "live" in this repo vs another.

I do think that it's important for edX's current developers that touch the front-end to define together what they think the standards for code quality, style, and guidelines should be. edx/edx-documentation#906 struck me more as an important "bird's eye view" intro on building edX UI. I see this as a separate concern; a quick reference to verify the quality of code changes.

style-guide.md
dan-f: (shrug) I've always used it. Typing `var` over and over gets annoying.

Then again, one doesn't need to squash all declarations into the same `var` statement in order to take advantage of hoisting. As long as they're all together at the top of the function.

Others?

style-guide.md
dan-f: Huh. My view was that modules are concerned by both syntax and semantics.

ES2015 defines both syntax _and_ semantics for imports. Edge supports it, but mosts others will need webpack. Nevertheless, this is distinct from AMD or CommonJS.

So I don't understand why we shouldn't mention modules here.

style-guide.md
dsjen: Very interesting, @bjacobel!

style-guide.md
```diff
@@ -0,0 +1,165 @@
...
+
+Use them!
+
+### Line Length
+
+Break lines at `120` columns.
+
+### Strings
+
+Use single quotes for string literals, unless the string itself contains single
```
dsjen: 👍

style-guide.md
dsjen: I am in the camp of one `var` declaration too, but open to other reasonings!  I think for me it's about convenience of not having to type `var` again and again, as well as additional `var`s standing out in case they've been declared w/in the function.

style-guide.md
```diff
@@ -0,0 +1,165 @@
...
+        throw new Error('No course specified');
+    }
+
+    courseModel.enroll(userModel);
+}
+```
+
+### Brackets, Braces, and Parentheses
+
+"Cuddle" the second keyword of `if/else`, `do/while`, and `try/catch` statements
```
dsjen: 👍

pdesjardins:
@dan-f I think this content is what we are trying to centralize in the language style guide section of the developer's guide. PR 906 adds two sections to the developer's guide, one is the bird's eye view and the other is the JavaScript style guide.

I suggest converting this PR into a developer's guide PR and merge the content with what we gathered in PR 906.

style-guide.md
ehteshamkafeel: @dan-f I have been a fan of one `var` at the top of function. But apart from hoisting and less typing there is no reason to use it. What we have learned from other languages is that variable declaration should be as close to it's use as possible, it makes the reading much easier. With ES6 when we will be writing `let` this pattern will go away I think. getify (you don't know js) has a very different opinion [https://twitter.com/getify/status/726917076812599296](https://twitter.com/getify/status/726917076812599296).

PS. while debugging all statements and expressions with one `var` evaluate at the same time.

dan-f:
If another edX JS developer is interested in taking over/organizing a developer-centric style guide, feel free to take over.

bjacobel:
@dan-f @pdesjardins I am! I'll close this for now and work on codifying it both in writing (as part of the developer's guide) and in our eslint config when Andy and I work on that next week.

PR #72, opened 2016-06-09
Add breadcrumbs JavaScript in preparation for new Pattern Library component

bjacobel:
Refs [TNL-4599](https://openedx.atlassian.net/browse/TNL-4599)

Adds backbone view, model and test for a breadcrumb component soon to be added to the UXPL.

[UXPL PR is here.](https://github.com/edx/ux-pattern-library/pull/334)

## Testing Checklist
- [x] Write accessibility (a11y) tests for all new UI.
- [x] Manually test responsive behavior.

## Non-testing Checklist
- [x] Consider any documentation your change might need, and which users will be affected by this change.
- [x] Consult with the Pattern Library team if adding a new component.

## Post-review

## Reviewers
- [x] @andy-armstrong 
- [ ] @dianakhuang 
- [x] @clrux (a11y review)

### DIFF
--- a/src/js/breadcrumbs/breadcrumbs-model.js
+++ b/src/js/breadcrumbs/breadcrumbs-model.js
@@ -1,0 +1,40 @@
+/**
+ * A Backbone model that works with the BreadcrumbsView to provide breadcrumb navigation
+ *
+ * Here's what initializing a BreadcrumbsModel looks like:
+ *
+ *~~~ javascript
+ * var model = new BreadcrumbsModel({
+ *     breadcrumbs: [
+ *         {
+ *             url: '/',
+ *             title: 'Item List'
+ *         },
+ *         {
+ *             url: '/details/1',
+ *             title: 'Item Details'
+ *         }
+ *     ],
+ *     label: 'Demo Page'
+ * });
+ *~~~
+ * @module BreadcrumbsModel
+ */
+;(function (define) {
+    'use strict';
+
+    define(['backbone'], function (Backbone) {
+        var BreadcrumbsModel = Backbone.Model.extend({
+            defaults: {
+                'breadcrumbs': null,
+                'label': ''
+            }
+        });
+
+        return BreadcrumbsModel;
+    });
+}).call(
+    this,
+    // Use the default 'define' function if available, else use 'RequireJS.define'
+    typeof define === 'function' && define.amd ? define : RequireJS.define
+);
--- a/src/js/breadcrumbs/breadcrumbs-view.js
+++ b/src/js/breadcrumbs/breadcrumbs-view.js
@@ -1,0 +1,46 @@
+/**
+ * A Backbone view that renders breadcrumbs-type tiered navigation.
+ *
+ * Initialize the view by passing in the following attributes:
+ *
+ *~~~ javascript
+ * var view = new DropdownMenuView({
+ *     el: $('selector for element that will contain breadcrumbs'),
+ *     model: new BreadcrumbsModel({
+ *         breadcrumbs: [{url: '/', title: 'Overview'}]
+ *     }),
+ *     events: {
+ *         'click nav.breadcrumbs a.nav-item': function (event) {
+ *             event.preventDefault();
+ *             window.location = $(event.currentTarget).attr('href');
+ *         }
+ *     }
+ * });
+ *~~~
+ * @module BreadcrumbsView
+ */
+;(function (define) {
+    'use strict';
+    define(['backbone', 'edx-ui-toolkit/js/utils/html-utils', 'text!./breadcrumbs.underscore'],
+        function (Backbone, HtmlUtils, breadcrumbsTemplate) {
+            var BreadcrumbsView = Backbone.View.extend({
+                initialize: function (options) {
+                    this.template = HtmlUtils.template(breadcrumbsTemplate);
+                    this.listenTo(this.model, 'change', this.render);
+                    this.render();
+                },
+
+                render: function () {
+                    var json = this.model.attributes;
+                    HtmlUtils.setHtml(this.$el, this.template(json));
+                    return this;
+                }
+            });
+
+            return BreadcrumbsView;
+        });
+}).call(
+    this,
+    // Use the default 'define' function if available, else use 'RequireJS.define'
+    typeof define === 'function' && define.amd ? define : RequireJS.define
+);
--- a/src/js/breadcrumbs/breadcrumbs.underscore
+++ b/src/js/breadcrumbs/breadcrumbs.underscore
@@ -1,0 +1,12 @@
+<div class="sr-is-focusable" tabindex="-1"></div>
+<% if (breadcrumbs !== null && breadcrumbs.length > 0) { %>
+    <nav class="breadcrumbs list-inline" aria-label="<%- label %>">
+        <% _.each(breadcrumbs.slice(0, -1), function (breadcrumb) { %>
+            <span class="nav-item">
+                <a href="<%- breadcrumb.url %>"><%- breadcrumb.title %></a>
+                <span class="fa fa-angle-right" aria-hidden="true"></span>
+            </span>
+        <% }) %>
+        <span class="nav-item"><%- breadcrumbs[breadcrumbs.length - 1].title %></span>
+    </nav>
+<% } %>
--- a/src/js/breadcrumbs/specs/breadcrumbs-spec.js
+++ b/src/js/breadcrumbs/specs/breadcrumbs-spec.js
@@ -1,0 +1,67 @@
+(function (define) {
+    'use strict';
+
+    define([
+            'jquery',
+            '../../utils/html-utils.js',
+            '../../utils/spec-helpers/spec-helpers.js',
+            '../breadcrumbs-view.js',
+            '../breadcrumbs-model.js'
+           ],
+           function($, HtmlUtils, SpecHelpers, BreadcrumbsView, BreadcrumbsModel) {
+
+               describe('BreadcrumbsView', function () {
+                   var model, view;
+
+                   beforeEach(function () {
+                       model = new BreadcrumbsModel();
+                       view = new BreadcrumbsView({
+                           model: model
+                       });
+                   });
+
+                   it('does not show breadcrumbs by default', function () {
+                       expect(view.$el.html()).not.toContain('<nav class="breadcrumbs">');
+                   });
+
+                   SpecHelpers.withData({
+                       'with no breadcrumbs': [[]],
+                       'with one breadcrumb': [[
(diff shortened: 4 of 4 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "edx-ui-toolkit",
-  "version": "1.1.0",
+  "version": "1.2.0",
```
andy-armstrong: Nit: we bump the version in a release PR so don't update it here yet.

src/js/breadcrumbs/breadcrumbs-model.js
```diff
@@ -0,0 +1,18 @@
+;(function (define) {
+    'use strict';
+
+    define(['backbone'], function (Backbone) {
+        var BreadcrumbsModel = Backbone.Model.extend({
+            defaults: {
+                'breadcrumbs': null,
+                'nav_aria_label': ''
```
andy-armstrong: Nit: I'd call this something more semantic, e.g. ``label`` or ``title``. It isn't necessarily only going to be used by ARIA.

src/js/breadcrumbs/breadcrumbs-view.js
```diff
@@ -0,0 +1,25 @@
+;(function (define) {
+    'use strict';
+    define(['backbone', 'underscore', 'text!./breadcrumbs.underscore'],
+        function (Backbone, _, breadcrumbsTemplate) {
+            var BreadcrumbsView = Backbone.View.extend({
+                initialize: function (options) {
+                    this.template = _.template(breadcrumbsTemplate);
```
andy-armstrong: Use ``HtmlUtils.template`` instead of ``_.template``.

src/js/breadcrumbs/breadcrumbs-view.js
```diff
@@ -0,0 +1,25 @@
...
+                    this.listenTo(this.model, 'change', this.render);
+                    this.render();
+                },
+
+                render: function () {
+                    var json = this.model.attributes;
+                    this.$el.html(this.template(json));
```
andy-armstrong: Use ``HtmlUtils.setHtml`` instead of ``$.html``. Note that we don't yet have safe template linting in this repository, but this would be flagged in edx-platform.

src/js/breadcrumbs/breadcrumbs.underscore
```diff
@@ -0,0 +1,9 @@
+<div class="sr-is-focusable" tabindex="-1"></div>
+<% if (breadcrumbs !== null && breadcrumbs.length > 0) { %>
+    <nav class="breadcrumbs" aria-label="<%- nav_aria_label %>">
+    <% _.each(breadcrumbs, function (breadcrumb) { %>
+        <a class="nav-item" href="<%= breadcrumb.url %>"><%- breadcrumb.title %></a>
+        <span class="icon icon-angle-right" aria-hidden="true"></span>
```
andy-armstrong: This will need to switch to a Font Awesome icon once @alisan617's PR lands.

src/js/breadcrumbs/breadcrumbs-model.js
```diff
@@ -0,0 +1,18 @@
+;(function (define) {
+    'use strict';
+
+    define(['backbone'], function (Backbone) {
+        var BreadcrumbsModel = Backbone.Model.extend({
```
andy-armstrong: Add some JSDoc so that your new files get documented.

src/js/breadcrumbs/breadcrumbs-view.js
```diff
@@ -0,0 +1,25 @@
+;(function (define) {
+    'use strict';
+    define(['backbone', 'underscore', 'text!./breadcrumbs.underscore'],
+        function (Backbone, _, breadcrumbsTemplate) {
+            var BreadcrumbsView = Backbone.View.extend({
```
andy-armstrong: Add some JSDoc so that your new files get documented.

src/js/breadcrumbs/specs/breadcrumbs-spec.js
```diff
@@ -0,0 +1,44 @@
...
+    'use strict';
+
+    define(['jquery',
+            'underscore',
+            '../breadcrumbs-view.js',
+            '../breadcrumbs-model.js'
+           ],
+           function($, _, BreadcrumbsView, BreadcrumbsModel) {
+
+               describe('Breadcrumbs component', function () {
```
andy-armstrong: Nit: I tend to name specs for the class, so this would be ``describe('BreadcrumbsView', ...``

src/js/breadcrumbs/specs/breadcrumbs-spec.js
```diff
@@ -0,0 +1,44 @@
...
+                       view = new BreadcrumbsView({
+                           model: model
+                       });
+                   });
+
+                   it('does not show breadcrumbs by default', function () {
+                       expect(view.$el.html()).not.toContain('<nav class="breadcrumbs">');
+                   });
+
+                   it('shows breadcrumbs if they are supplied', function () {
```
andy-armstrong: It would be good to test a variety of different breadcrumbs. Look at using the ``SpecHelpers.withData`` function so that you can do this in a declarative way, e.g. https://github.com/edx/edx-ui-toolkit/blob/master/src/js/utils/specs/html-utils-spec.js#L101

Some tests I would include:
 - rendering a single breadcrumb
 - rendering two items in a breadcrumb
 - rendering Unicode characters
 - rendering labels that include HTML to verify that they are safely escaped

src/js/breadcrumbs/breadcrumbs.underscore
```diff
@@ -0,0 +1,9 @@
+<div class="sr-is-focusable" tabindex="-1"></div>
+<% if (breadcrumbs !== null && breadcrumbs.length > 0) { %>
+    <nav class="breadcrumbs" aria-label="<%- nav_aria_label %>">
+    <% _.each(breadcrumbs, function (breadcrumb) { %>
+        <a class="nav-item" href="<%= breadcrumb.url %>"><%- breadcrumb.title %></a>
```
andy-armstrong: Our best practice is to use ``<%-`` even for URLs (the safe template linter would complain about this).

package.json
andy-armstrong: See here for details: https://openedx.atlassian.net/wiki/display/FEDX/UI+Toolkit+release+process

src/js/breadcrumbs/specs/breadcrumbs-spec.js
```diff
@@ -0,0 +1,44 @@
...
+                       testBreadcrumbs([
```
andy-armstrong: Nit: I like to have the except calls explicitly in the body of the test so that it is easy to read what is being tested. If you use ``SpecHelpers.withData`` then you won't need the ``testBreadcrumbs`` function which will naturally make this happen.

andy-armstrong:
@bjacobel This is looking great. I have a number of minor comments so let me know when you're ready for a second review. When you've added documentation you should use ``gulp preview`` to generate a doc preview for reviewers to look at too.

src/js/breadcrumbs/specs/breadcrumbs-spec.js
andy-armstrong: I just realized that it is legitimate to create a BreadcrumbView with no items, so I'd add that too.

src/js/breadcrumbs/breadcrumbs-view.js
```diff
@@ -0,0 +1,25 @@
...
+                initialize: function (options) {
+                    this.template = _.template(breadcrumbsTemplate);
+                    this.listenTo(this.model, 'change', this.render);
+                    this.render();
```
andy-armstrong: I prefer that views do not call ``render`` in their initialize methods. There are a number of reasons for this:
 - it makes it hard to control when the view is rendered
 - it means that subclasses will render before their own initialization has completed
 - it will render twice if you create a view, configure it, add children etc
 - it doesn't allow for a component to be rendered server-side

bjacobel:
Ready for re-review.

src/js/breadcrumbs/breadcrumbs-view.js
```diff
@@ -1,17 +1,38 @@
...
+ *     }
+ * });
+ *~~~
+ * @module BreadcrumbsView
+ */
 ;(function (define) {
     'use strict';
-    define(['backbone', 'underscore', 'text!./breadcrumbs.underscore'],
-        function (Backbone, _, breadcrumbsTemplate) {
+    define(['backbone', '../utils/html-utils.js', 'text!./breadcrumbs.underscore'],
```
andy-armstrong: I had trouble consuming the UI Toolkit in edx-platform when using relative paths. I think you might need to use a full path like this:

https://github.com/edx/edx-ui-toolkit/blob/master/src/js/utils/html-utils.js#L14

andy-armstrong:
👍 once you fix the import to not use a relative path.

bjacobel:
@andy-armstrong Sorry, which import?

src/js/breadcrumbs/breadcrumbs-view.js
```diff
@@ -0,0 +1,46 @@
...
+ *             window.location = $(event.currentTarget).attr('href');
+ *         }
+ *     }
+ * });
+ *~~~
+ * @module BreadcrumbsView
+ */
+;(function (define) {
+    'use strict';
+    define(['backbone', '../utils/html-utils.js', 'text!./breadcrumbs.underscore'],
```
andy-armstrong: @bjacobel I don't know where my comment went, but I had trouble using .. paths when consuming the UI Toolkit in edx-platform. I think this needs to be ``edx-ui-toolkit/js/utils/html-utils.js``. You can see in html-utils itself that we do this to access string-utils.

andy-armstrong: @bjacobel This is what I meant by import, to be clear.

bjacobel: Sounds good. Should be fine in the tests, though, right?

bjacobel:
tagging @clrux for a11y review here, thanks Chris!

clrux:
From code alone, this looks fine. Is there a way to test?

bjacobel:
@clrux There's an example of it in the preview docs site I built for the [UXPL PR](https://github.com/edx/ux-pattern-library/pull/334), although it currently doesn't point to the latest version of this code - I'll update the preview site there once I merge this and ship a new release to NPM.

bjacobel:
Updated with more a11y fixes with help from @clrux. Chris, you want to give a final thumbs?

clrux:
Just get the styling fixed up. Markup and feedback sound OK! 👍

PR #75, opened 2016-06-20
Implement ESLint with eslint-config-edx

bjacobel:
Install ESLint, and adopt [our new styleguide](https://github.com/bjacobel/eslint-config-edx).

~~If you have comments about the ESLint rules that are responsible for suggesting these changes, you can make them here: https://github.com/bjacobel/eslint-config-edx/pull/1. The ruleset/styleguide is not released/finalized yet, but will be 1.0 soon, so hurry!~~ `eslint-config-edx` is now 1.0 [on NPM](https://npmjs.com/package/eslint-config-edx)

## Testing Checklist
N/A

## Non-testing Checklist
- [ ] Consider any documentation your change might need, and which users will be affected by this change.

## Post-review

- [x] :rotating_light: TODO before merge: update package.json version of eslint-config-edx to NPM 1.0 :rotating_light:

## Reviewers
- [x] @andy-armstrong 
- [x] @alisan617 
- [x] @AlasdairSwan 
- [x] @dsjen 


gulp/tasks/lint.js
```diff
@@ -1,21 +1,21 @@
 (function() {
     'use strict';
 
-    var jshint = require('gulp-jshint'),
+    var eslint = require('gulp-eslint'),
         gulp = require('gulp'),
         paths = {
             lint: [
```
andy-armstrong: Do we need both this and the .eslintignore? Maybe this can just be ``**/*.js`` and let the exceptions be handled in the ignore file.

src/js/disclosure/disclosure-view.js
```diff
@@ -8,7 +8,7 @@
  *
  * @module DisclosureView
  */
-;(function(define) {
+(function(define) {
```
andy-armstrong: Is the starting semicolon not allowed? I think we use this extensively in edx-platform, but maybe this is a good incentive for us to remove it.

src/js/dropdown-menu/specs/dropdown-menu-view-spec.js
```diff
@@ -256,21 +257,33 @@ define([
                     closeOnKeypressTest(constants.keyCodes.space);
                 });
 
-                it('should return to the button after pressing down arrow key while at bottom of dropdown menu', function() {
-                    focusTrapDown(constants.keyCodes.down, 4);
-                });
+                it(
```
andy-armstrong: Are these lines too long? My preference is to have a shorter message, and keep everything on one line. I like tests to read concisely.

andy-armstrong:
@bjacobel It looks like you have a leftover call to the old jscs gulp task.

src/js/dropdown-menu/specs/dropdown-menu-view-spec.js
bjacobel: Yeah, 120 is our max and these were too long. We can fix it by just shortening the message as well.

src/js/disclosure/disclosure-view.js
bjacobel: Yeah, I'm not sure what rule is doing this, but it's not allowed. Couple of things we can do:

- allow it explicitly in the eslint-config-edx
- allow it with an override only in places we're already doing this (edx-platform, edx-ui-toolkit)
- make sure it's not necessary any of the places we do it, and remove them all
- make the r.js optimizer or webpack do this when it concatenates them, and remove them all

Personally I'd like to get rid of them all, as I understand it it's a practice that only has a benefit when you're already doing something bad and we should just be fixing the bad code instead of allowing it to limp along :)

andy-armstrong: Agreed. Let's leave the rule as is and just clean up the code. I'm fairly sure that r.js already handles this, so the remaining question is if any of our legacy code that bundles with Django staticfiles will be a problem. I doubt it, and as you say this is a very unlikely issue (where a page loads incomplete JS that can be fixed by a single semicolon).

gulp/tasks/doc.js
```diff
@@ -6,113 +6,116 @@
...
-gulp.task('doc-publish', ['clean', 'doc-build'], function() {
-    return gulp.src(config.gitHubPages.files)
-        .pipe(ghPages());
-});
+    ghPages = require('gulp-gh-pages');
+
+(function() {
+    'use strict';
+
+    var renameAsMarkdown, generateDocFor;
```
dsjen: I guess the convention has been to declare variables in alphabetical order unless declarations get messed up.

test/karma.conf.js
```diff
@@ -1,10 +1,12 @@
...
 
-module.exports = function(config, options) {
+module.exports = function(config, overrideOptions) {
     'use strict';
 
+    var options;
+
     // If not being overridden, then specify the default headless configuration
-    if (!options) {
+    if (!overrideOptions) {
```
dsjen: Nit: I completely realize you're just trying to get the styles to pass, so please don't feel obligated to implement this suggestion.  I'd consider changing this `if` statement to the positive: 
```
if (overrideOptions) {
    options = overrideOptions;
} else {
...
}
```

Also, couldn't we do something like [defaults](http://underscorejs.org/#defaults)?

test/require-config.js
```diff
@@ -6,30 +6,31 @@ require.config({
     baseUrl: '/base',
     waitSeconds: 60,
     paths: {
-        backbone: 'node_modules/backbone/backbone',
+        'backbone': 'node_modules/backbone/backbone',
```
dsjen: I thought the linter didn't recommend quotes around fields. No?

andy-armstrong: Interesting. I don't like quotes around fields either. ESLint has a huge number of configuration options for this rule:

http://eslint.org/docs/rules/quote-props

I think I like the option ``"consistent-as-needed"``, but I'd also be fine with ``"as-needed"`` which is the more lenient version.

dsjen: Yes, please! 💯

bjacobel: We're currently set to `consistent`, but I got confused and actually documented the `consistent-as-needed` behavior I think? So maybe that's what we should go with because that's what people already thumbsed up on that PR.

(Here, I think the linter is tripping on `backbone.paginator`, which isn't a valid property name without quotes, so they're all being quoted. But if `backbone.paginator` wasn't here, and we were using `consistent-as-needed`, none of these would need quotes.)

dsjen: Seeing the rules applied to a real project is super helpful and is helping flesh this out.  Thank you!

package.json
```diff
@@ -31,21 +34,18 @@
     "css-loader": "~0.23.1",
     "del": "~2.2.0",
     "edx-pattern-library": "~0.12.5",
+    "eslint-config-edx": "bjacobel/eslint-config-edx#bjacobel/feedback",
```
alisan617: Isn't this supposed to be edx/eslint-config-edx 0.0.1 ?

bjacobel: The `0.0.1` that's on NPM right now was released last Friday so it's a little stale, but I'm going to release a `1.0.0` to NPM after https://github.com/edx/eslint-config-edx/pull/1 merges and then this can be updated to point there.

bjacobel:
This has no more linter issues and is ready for review.

(There's some issue by which the coverage percentage dropped... I'll look into that.)

gulp/tasks/doc.js
bjacobel: I don't see us following this convention anywhere very frequently. Is it something you feel strongly about starting to enforce? We can look into if there's an eslint rule for it.

bjacobel:
[MRW Coveralls is failing on a 0.02% decrease, but none of the files on Coveralls actually show a decrease in lines covered](https://gifs.bjacobel.com/ragequit.gif)

gulp/tasks/doc.js
andy-armstrong: I wasn't aware of that convention. I tend to declare them in the order I'm going to use them.

src/js/utils/spec-helpers/spec-helpers.js
```diff
@@ -15,17 +17,12 @@ define([], function() {
...
-            if (data.hasOwnProperty(name)) {
-                (function(name) {
-                    it(name, function() {
-                        func.apply(this, data[name]);
-                    });
-                })(name);
-            }
-        }
+    withData = function(data, func) {
+        Object.keys(data).forEach(function(key) {
```
bjacobel: Note to reviewers: `for (... in ...)` is banned by Airbnb ESLint because a) it has performance issues in some browsers and b) it loops over _all_ enumerable properties, not just object keys (e.g., `toString`)

In ES6 we can use [`for (... of ...)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) to do what `for (... in ...)` _looks_ like it does.

gulp/tasks/doc.js
alisan617: I tend to declare them in the order I'm going to use, too.

bjacobel:
[this coveralls build](https://coveralls.io/builds/6770596): 883 of 935 relevant lines covered
[latest master build](https://coveralls.io/builds/6743333): 886 of 938 relevant lines covered

So there's three less lines covered, and three less lines overall... according to Coveralls that's a build failure. Wat.

bjacobel:
Hi folks! Would love to get this merged in this week, if we can - that's when I said we'd be turning it on in my email to edx-code.

@dsjen and @andy-armstrong have already done an initial review, mind giving final signoff if I've addressed all comments? @AlasdairSwan and @alisan617, would love to hear your feedback too.

Haven't figured out a good way to address the coveralls failure yet. Wish there was something analogous to `[skip ci]` commit comments to skip a Travis build. Any ideas?

alisan617:
LGTM 👍

gulp/tasks/doc.js
AlasdairSwan: I'm with @andy-armstrong and @alisan617, I declare them in the order used. Alphabetical could be complicated if var aardvark relies on var zoo.

gulp/.eslintrc.json
```diff
@@ -0,0 +1,5 @@
+{
+    "rules": {
+        "no-console": "off"
```
AlasdairSwan: @bjacobel I quite like this rule, forcing people to merge cleaner code. Is there a reason we would want ```console.log``` statements in the codebase? I would be fine keeping ```warn``` or ```error``` if it is needed but vote to get rid of debugging statements from the code.

gulp/tasks/doc.js
```diff
@@ -6,113 +6,116 @@
...
+
+    renameAsMarkdown = function(path) {
+        var renamedPath = path;
+        renamedPath.extname = '.md';
+        return renamedPath;
+    };
+
+    generateDocFor = function(options) {
+        var i, sources;
+        for (i = 0; i < options.sources.length; i++) {
```
AlasdairSwan: Nitpick: I would prefer that the length is calculated before the loop and stored as a variable so that it doesn't have to be calculated with each iteration.

gulp/.eslintrc.json
bjacobel: ESLint allows per-directory configs, so this only applies to scripts in the `gulp` folder. I decided to allow `console` in Gulp scripts because 1) `console` is always defined in Node, whereas one of the reasons the `no-console` rule exists is IE9 doesn't implement `console`, and 2) `console.log` is pretty common in our gulp scripts as a means of feedback

src/js/pagination/specs/paging-collection-spec.js
```diff
@@ -345,25 +350,27 @@ define(['jquery',
 
                     it('triggers an error event when the requested page is out of range',
                         AjaxHelpers.withFakeRequests(function(requests) {
-                            var errorTriggered = false;
+                            // TODO: Re-enable expectations and remove eslint ignores
```
AlasdairSwan: Is this TODO going to be done in a later PR or this one?

test/require-config.js
AlasdairSwan: So has this change to ```consistent-as-needed```? And if so doesn't that mean that quotes can be removed when not needed (i.e. everything apart from 'backbone.paginator'?

gulp/.eslintrc.json
AlasdairSwan: Cool, thanks for the explanation.

src/js/pagination/specs/paging-collection-spec.js
bjacobel: Probably a later PR. There's some sort of issue with AjaxHelpers that's been around forever. I tried debugging it when it came up in this PR (because `errorTriggered` was declared but never used, with those tests commented) but couldn't figure it out easily.

@andy-armstrong, I think the AjaxHelpers issues with these tests might actually be related to the other weird $.ajax thing I showed you the other day.

test/require-config.js
bjacobel: yep, we're using `consistent-as-needed` now, which means that because backbone.paginator needs quotes, all other properties of this object do too.

AlasdairSwan:
👍

test/require-config.js
dsjen: Oh...  Is there a way to get rid of that?  It seems restrictive to require quotes on fields because there happens to be a field that requires quotes.

bjacobel: Sure. The rule is [`quote-props`](http://eslint.org/docs/rules/quote-props), if you're interested in opening up a discussion about it, maybe file an issue to the eslint-config-edx repo and tag people?

(I have Opinions about why `consistent-as-needed` is the right way to do object property quoting but I think they should be aired over on that repo, not in this PR)

dsjen: Thanks, can't wait!  https://github.com/edx/eslint-config-edx/issues/10

dsjen:
👍

andy-armstrong:
👍

src/js/pagination/specs/paging-collection-spec.js
bjacobel: I fixed it! There's a bad bug in `sinon@1.17.4`, downgrading it to the previous version allows these tests to be uncommented and pass. See https://github.com/sinonjs/sinon/issues/1040

Also fixes the coverage issues.

😌

andy-armstrong:
@bjacobel Awesome find with the sinon bug! Great coverage improvements too.

PR #78, opened 2016-06-21
Fix BreadcrumbsView docs that still said DropdownMenuView

bjacobel:
@andy-armstrong 

http://ux-test.edx.org/bjacobel/fix-docs-copypasta/views/breadcrumbs/breadcrumbs-view/

baboonlaptop


andy-armstrong:
👍

PR #79, opened 2016-06-21
Add some configuration for mention-bot

bjacobel:
@dan-f asked me if we could find a way to keep @mention-bot to not ping him on PRs for code review anymore. I've also seen it add Brian to some reviews so I added him to the exclude list as well.

Also added option `skipAlreadyMentionedPR`, which will prevent @-msging people who have already been explicitly listed for review.

- [ ] @andy-armstrong

### DIFF
--- a/.mention-bot
+++ b/.mention-bot
@@ -1,7 +1,8 @@
 {
   "userBlacklist": [
     "dan-f",
-    "talbs"
+    "talbs",
+    "peter-fogg"
   ],
   "skipAlreadyMentionedPR": true
 }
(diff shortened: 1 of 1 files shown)


dan-f:
Thank you :)

andy-armstrong:
Nice idea 👍 Do you want to do the same for the Pattern Library? Also @peter-fogg may want to be removed too.

peter-fogg:
Thanks @andy-armstrong and @bjacobel -- my inbox can be a bit happier now 💃

PR #82, opened 2016-07-01
Write specs for ajax helpers

bjacobel:
## Description

Increase test coverage for `AjaxHelpers`, our lowest coverage file. Recently we had some mysterious failing tests when Sinon broke the `fakeServer` in an upgrade, more test coverage here will help diagnose issues like that easily.

## Manual Test Plan
N/A

## Testing Checklist
- [x] Write unit tests ~~for all new features~~.
- ~~Write accessibility (a11y) tests for all new UI.~~ N/A
- ~~Manually test responsive behavior.~~ N/A

## Non-testing Checklist
N/A

## Post-review

## Reviewers
- [x] @andy-armstrong
- [ ] @dsjen

FYI @andy-armstrong 

### DIFF
--- a/src/js/utils/specs/ajax-helpers-spec.js
+++ b/src/js/utils/specs/ajax-helpers-spec.js
@@ -1,0 +1,185 @@
+define(
+    [
+        'jquery',
+        'sinon',
+        '../../utils/spec-helpers/ajax-helpers.js'
+    ],
+    function($, sinon, AjaxHelpers) {
+        'use strict';
+
+        describe('AjaxHelpers', function() {
+            var testUrl = 'https://example.com',
+                testData = {foo: 'bar', fizz: 'buzz'},
+                testQuerystring = $.param(testData);
+
+            describe('fakeServer', function() {
+                it('responds with the contents of the first argument', function() {
+                    var ajaxSpy = jasmine.createSpy('ajaxSpy');
+                    var server = AjaxHelpers.server('foo');
+
+                    $.ajax(testUrl).done(ajaxSpy);
+                    server.respond();
+
+                    expect(ajaxSpy).toHaveBeenCalledWith('foo', 'success', jasmine.any(Object));
+                });
+            });
+
+            describe('fakeRequests', function() {
+                it('returns an array that tracks fake requests', function() {
+                    // We're really already doing this already in all the tests that call AjaxHelpers.withFakeRequests
+                    // down below, but Istanbul doesn't pick those up because they're not "in" a spec...
+                    // so do a quick manual test.
+                    var requests = AjaxHelpers.requests();
+
+                    $.ajax(testUrl);
+
+                    expect(requests instanceof Array).toBeTruthy();
+                    expect(requests.hasOwnProperty('currentIndex')).toBeTruthy();
+                    AjaxHelpers.expectRequest(requests, 'GET', testUrl);
+                });
+            });
+
+            describe('request helper', function() {
+                describe('expectRequest', function() {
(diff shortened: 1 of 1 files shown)


andy-armstrong:
👍 Thanks for the additional coverage!

src/js/utils/specs/ajax-helpers-spec.js
```diff
@@ -0,0 +1,166 @@
...
+    ],
+    function($, sinon, AjaxHelpers) {
+        'use strict';
+
+        describe('AjaxHelpers', function() {
+            var testUrl = 'https://example.com',
+                testData = {foo: 'bar', fizz: 'buzz'},
+                testQuerystring = $.param(testData);
+
+            describe('fakeServer', function() {
```
dsjen: Are these empty on purpose?

src/js/utils/specs/ajax-helpers-spec.js
```diff
@@ -0,0 +1,166 @@
...
+                            data: testQuerystring
+                        });
+                        AjaxHelpers.expectPostRequest(requests, testUrl, testQuerystring);
+                    }));
+                });
+
+                describe('skipResetRequest', function() {
+                    it('verifies request was cancelled', AjaxHelpers.withFakeRequests(function(requests) {
+                        $.ajax(testUrl).abort();
+                        AjaxHelpers.skipResetRequest(requests);
```
dsjen: Nit: I'm not sure there's anything to do really, but I find it a bit odd to have a spec w/o and expects.  Maybe it's just the naming of `skipResetRequest` as it does check the `readyState`.  I guess one could check that `currentIndex` in incremented.

src/js/utils/specs/ajax-helpers-spec.js
bjacobel: I initially planned on writing tests for these two methods but ran out of time (and/or forgot 😅). They should still probably be tested at some point though. I can:

- delete these describe blocks
- mark them up with a `TODO: implement tests` and set them up to be skip for now with `xdescribe`
- finish writing the tests

Which do you think is best?

src/js/utils/specs/ajax-helpers-spec.js
bjacobel: I agree, I'm not a bug fan of the expectation helper pattern for this reason (because it makes it hard to see what's happening in tests). I'll add a line that checks to see if currentIndex is incremented, though, that's a good idea.

bjacobel: although... most of the `describe('request helper',` specs follow this patters and don't have any `expects` in them... so I'd have to add a currentIndex expect to each one... which sounds to me like something that should go in a helper function to maintain DRY... aaand we're back at square 1. Thoughts?

src/js/utils/specs/ajax-helpers-spec.js
dsjen: Ah, ok!  I don't have strong opinions, but find personally, TODOs sometimes sit around for a while and I just end up feeling guilty about it.  Maybe delete and add a story for adding the tests in?

src/js/utils/specs/ajax-helpers-spec.js
dsjen: Yuck!  I guess perhaps there's nothing to do here then.  Thanks for clarifying!  Maybe adding a comment would be helpful in understanding what's going on.

bjacobel:
@dsjen Just coming back to this now. Pushed new tests for those two previously empty `describe` blocks that should bring the file I'm testing up to 100% coverage.

andy-armstrong:
👍

PR #83, opened 2016-08-04
Update to newest eslint-config-edx. Remove IIFEs in Node code.

bjacobel:
Update to the version of eslint-config-edx that contains fixes to 'use strict'; rule. Node code doesn't need to be wrapped in IIFEs and should 'use strict'; at global scope.

(This undoes all the wrapping of this code I thought I needed to do a few weeks ago....)

- [ ] @andy-armstrong 
- [x] @alisan617

### DIFF
--- a/gulp/tasks/clean.js
+++ b/gulp/tasks/clean.js
@@ -1,21 +1,20 @@
+'use strict';
+
 var gulp = require('gulp'),
     config = require('../config'),
     del = require('del');
 
-(function() {
-    'use strict';
-    gulp.task('clean', function() {
-        return del([
-            // Remove the Jekyll site
-            config.documentation.targetDir,
+gulp.task('clean', function() {
+    return del([
+        // Remove the Jekyll site
+        config.documentation.targetDir,
 
-            // Remove the preview site
-            config.documentation.previewTargetDir,
+        // Remove the preview site
+        config.documentation.previewTargetDir,
 
-            // Remove the JSDoc generated markdown
-            config.documentation.testing.output,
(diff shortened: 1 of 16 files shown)


gulp/tasks/clean.js
```diff
@@ -1,21 +1,21 @@
+/* eslint-env node */
```
andy-armstrong: I thought you had a rule so that the gulp directory was automatically treated as being a node environment?

andy-armstrong:
👍

gulp/tasks/clean.js
bjacobel: I thought I could just do 

`./gulp/.eslintrc.json`
```
{
  "env": {
    "node: true
  }
}
```

but that wasn't working for some reason. I'll take another look at it before I merge this. Got sidetracked when my jenkins build finished :)

bjacobel: Updated to fix this, the folder-level .eslintrc.json now has the right configuration to make this folder pick up the right setting for the `strict` rule.

bjacobel:
Ready for re-review - @alisan617, if you could when you get a chance? Thanks!

package.json
```diff
@@ -34,7 +34,7 @@
     "del": "~2.2.0",
     "edx-pattern-library": "~0.12.5",
```
alisan617: any reason why UXPL is not updated to the latest?

bjacobel: We generally don't update deps until we need to. Nothing visual has changed in the UITK for a while.

PR #84, opened 2016-08-23
Add an OEP-2 compliant openedx.yaml file

cpennington:
This adds an `openedx.yaml` file, as described by OEP-2:
http://open-edx-proposals.readthedocs.io/en/latest/oeps/oep-0002.html

The data in this file was transformed from the contents of
edx/repo-tools-data:repos.yaml

### DIFF
--- a/openedx.yaml
+++ b/openedx.yaml
@@ -4,5 +4,5 @@
 nick: uitk
 oeps: {}
 openedx-release: {ref: master}
-owner: edx/edx-ux-team
+owner: edx/fedX-team
 track-pulls: true
--- a/openedx.yaml
+++ b/openedx.yaml
@@ -1,0 +1,8 @@
+# This file describes this Open edX repo, as described in OEP-2:
+# http://open-edx-proposals.readthedocs.io/en/latest/oeps/oep-0002.html#specification
+
+nick: uitk
+oeps: {}
+openedx-release: {ref: master}
+owner: MUST FILL IN OWNER
+track-pulls: true
--- a/openedx.yaml
+++ b/openedx.yaml
@@ -4,5 +4,5 @@
 nick: uitk
 oeps: {}
 openedx-release: {ref: master}
-owner: MUST FILL IN OWNER
+owner: edx/edx-ux-team
 track-pulls: true


cpennington:
@edx/edx-ux-team: I'm assigning this repo to you.

cpennington:
@edx/edx-ux-team: Any objections to me merging this?

bjacobel:
@cpennington I don't think UX is the right owner for this repo, it's mostly Javascript utils and component logic

cpennington:
@bjacobel: Who, then, do you think? TNL? Do we need a fedx team?

openedx.yaml
```diff
@@ -0,0 +1,8 @@
+# This file describes this Open edX repo, as described in OEP-2:
+# http://open-edx-proposals.readthedocs.io/en/latest/oeps/oep-0002.html#specification
+
+nick: uitk
+oeps: {}
+openedx-release: {ref: master}
+owner: edx/edx-ux-team
```
bjacobel: You could make an @edx/fedx alias that points to me, Alasdair, Alisan, Andy, Dennis and Tyler

bjacobel:
@cpennington Yep, I think so.

https://github.com/edx/edx-ui-toolkit/graphs/contributors

clrux:
Personally, I think FedX should focus on the Toolkit, and UX should focus on the UXPL. I doubt the UX team can actually build components so they'd definitely need assistance.

clrux:
But on that same note, the goal here (as with the pattern library) is consistency. If everybody is buildin' things willy nilly, there's likely not going to be consistency.

chris-mike:
As for consistency, the first thing that stands out is the font. We use open sans, you use something completely different. However, there's nothing here that's so obviously edx (no logo, etc.), so I wonder if it's even w/in scope of our normal review.    

@roderickmorales is the ux point person for the pattern library. He veers into fedex than the rest of the ux team.  Still, I think @bjacobel and @clrux  make valid points about what is expected for this particular review.

cpennington:
@bjacobel: I'm going to open a ticket w/o IT to create @edx/fedX-team, and then get @andy-armstrong to fill in the members.

andy-armstrong:
👍  thanks for making this happen @cpennington.

PR #85, opened 2016-09-28
DateUtil, moment req, tests

yro:
## Description
The intention of this PR is to get some early code review before roping in documentation and product. A formal PR with the requisite requirements below all seen to will follow.

## [TNL-5565](https://openedx.atlassian.net/browse/TNL-5565)

This add a language/timezone aware datetime formatting utility.

## Manual Test Plan
None that I am aware of

## Testing Checklist
- [x] Write accessibility (a11y) tests for all new UI.
- [x] Manually test responsive behavior.

## Non-testing Checklist
- [x] Consider any documentation your change might need, and which users will be affected by this change.
- [x] Consult with the Pattern Library team if adding a new component.

## Post-review

## Reviewers
TODO: In this section, tag specific reviewers you know will want to have a look at this PR. You can use the checklist below for organization:

- [x] @nasthagiri 
- [x] @efischer19 

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,8 @@
     "backbone.paginator": "~2.0.3",
     "gulp-shell": "^0.5.2",
     "jquery": "~2.2.0",
+    "moment": "^2.15.1",
+    "moment-timezone": "~0.5.5",
     "requirejs": "~2.1.22",
     "requirejs-text": "~2.0.12",
     "sinon": "1.17.3 || >1.17.4",
--- a/src/js/utils/date-utils.js
+++ b/src/js/utils/date-utils.js
@@ -1,0 +1,255 @@
+/**
+ * Create a formatted, locale-aware string representation of a datetime.
+ *
+ * This module contains general purpose functions for localizing and displaying
+ * date-times by timezone, locale and preferred language."
+ *
+ * Most of the heavy lifting is done by the {@link http://momentjs.com/ |'moments' js library}
+ * with the moment-with-locales and moment-timezones add-ons.
+ *
+ *
+ *
+ * Example:
+ *
+ *~~~ javascript
+ * var context = {
+ *    datetime: '2016-10-14 08:00:00+00:00',
+ *    timezone: 'Pacific/Kiritimati',
+ *    language: 'ar',
+ *    format: DateUtils.dateFormats.shortDate
+ * };
+ * localized_date = DateUtils.localize(context);
+ *~~~
+ *
+ * @module DateUtils
+ */
+
+
+(function(define) {
+    'use strict';
+    define([
+        'jquery',
+        'moment',
+        'moment-timezone'
+    ], function(
+        $,
+        moment,
+        momentTZ
+        ) {
+        var DateUtils,
+            localize,
+            stringToMoment,
+            localizeTime,
+            getSysTimezone,
+            getDisplayTime,
+            getDisplayLang,
+            getDisplayFormat,
+            isValid;
+
+        var dateFormatEnum = Object.freeze({
+            shortDate: 'll', // example: Jan 01, 2016
+            longDate: 'LLLL', // example: Friday, January 01, 2016
+            time: 'LTS z', // example: Jan 01, 2016 15:30 UTC
+            defaultFormat: 'll HH[:]mm z' // example: Jan 01, 2016 15:30 UTC
+        });
+
+        var DEFAULT_LANG = 'en';
+        var DEFAULT_TIMEZONE = 'UTC';
+        var DEFAULT_FORMAT = dateFormatEnum.defaultFormat;
+
+        /**
+         *
+         * Renders a localized date/time in a standard format.
+         *
+         * @param context
+         *
+         * - datetime: {string} The (UTC) datetime server-side string.
+         * - timezone: {string} (optional) the user set time zone preference, or 'None' if unset.
+         * This should be in a format like 'Region/Locale', e.g. 'Atlantic/Azores'. Defaults to 'UTC'.
+         * - language: {string} (optional), the user set language, an ISO 639-1 language string, defaults to 'en'
+         * - format: {object} (optional), a format constant. Defaults to DEFAULT_FORMAT.
+         *
+         * @returns {string} A formatted, timezone-offset, internationalized date and time.
+         *
+         */
+        localize = function(context) {
+            var localTime;
+            var displayTime = '';
+
+            if (isValid(context.datetime)) {
+                localTime = localizeTime(
+                    DateUtils.stringToMoment(context.datetime),
+                    context.timezone
+                );
+                displayTime = getDisplayTime(localTime, context.language, context.format);
+            }
+            return displayTime;
+        };
+
+        /**
+         * Converts a string to mutable 'moment' datetime object.
+         *
+         * The input datetimeString can be just a date, a date and time,
+         * or a complex date/time string with TZ offset. This is fairly wide-open, as
+         * many strings will satisfy this requirement.
+         * (e.g. 01-01-16 and Jan 1, 2016, and 2016-01-01 01:00:00+00:00 will work)
+         *
+         * The recommendation is to use a fully informed string, e.g. '2016-01-01 01:00:00+00:00',
+         * to limit possible date, time, and timezone ambiguities.
+         *
+         * @param {string} datetimeString. A string representation of a datetime.
+         * @returns {object} A moment.js UTC datetime object.
+         *
+        */
+        stringToMoment = function(datetimeString) {
+            var utcDateObject;
+            if (isValid(datetimeString)) {
+                utcDateObject = moment(datetimeString).utc();
+            } else {
+                utcDateObject = undefined;
+            }
+            return utcDateObject;
+        };
+
+        /**
+         *
+         * Returns a new time that is offset by the requested timezone.
+         *
+         * @param {Object} dateTimeObject A moment.js datetime object.
+         * @param {string} timezone A timezone representation, e.g. 'America/New_York'.
+         * @return {object} A moment.js datetime object in a determined timezone.
+         */
+        localizeTime = function(dateTimeObject, timezone) {
+            var localTime;
+            if (isValid(dateTimeObject) && isValid(timezone)) {
+                localTime = dateTimeObject.tz(timezone);
+            } else if (isValid(dateTimeObject)) {
+                localTime = dateTimeObject.tz(getSysTimezone());
+            } else {
+                localTime = '';
+            }
+            return localTime;
+        };
+
+        /**
+         *
+         * Helper function to determine system timezone, and fallback to UTC
+         * Determines timezone from argument, falling back to system/browser settings, then UTC.
+         *
+         * @return {string} Will attempt to return a determined timezone string, with a fallback to UTC.
+         *
+         */
+        getSysTimezone = function() {
+            var timezone = momentTZ.tz.guess();
+            if (!isValid(timezone)) {
+                timezone = DEFAULT_TIMEZONE;
+            }
+            return timezone;
+        };
+
+        /**
+         *
+         * Generate a timezone-aware and language-aware string.
+         *
+         * This includes a param for standard formatting constants.
+         *
+         * Browser preferred language is currently overridden by platform-set language preference.
+         * If no preference is set, the function attempts to determine the locale settings
+         * (e.g. en-US) from the browser preferences.
+         *
+         * This function will default to UTC/en-US display in default format in the case of
+         * an unresolvable language or timezone
+         *
+         * preferred_language = 'en' yields Oct. 14, 2016 09:00 BST
+         *
+         * preferred_language = 'ru' yields 14 окт 2016 г. 09:00 BST
+         *
+         * @param {object} localTime A moment.js datetime object offset to a determined timezone.
+         * @param {string} language an ISO 639-1 language code (e.g. 'en-US', 'fr')
+         * @param {object} format (optional) A desired date format, via the dateFormatEnum.
+         * @return {string} A language/timezone aware localized string
+        */
+        getDisplayTime = function(localTime, language, format) {
+            var displayTime = '';
+            var displayLang = getDisplayLang(language);
+            var displayFormat = getDisplayFormat(format);
+
+            if (isValid(displayLang) && isValid(displayFormat)) {
+                if (moment(localTime).isValid()) {
+                    displayTime = localTime.locale(displayLang).format(displayFormat);
+                }
+            }
+            return displayTime;
+        };
+
+        /**
+         * Helper function to determine sys/browser language (if unpassed)
+         *
+         * @param {string} language (optional) an ISO 639-1 language code (e.g. 'en-US', 'fr')
+         * @returns {string} an ISO 639-1 language code (e.g. 'en-US', 'fr')
+         */
+        getDisplayLang = function(language) {
+            var displayLang;
+            if (isValid(language)) {
+                displayLang = language;
+            } else {
+                displayLang = window.navigator.userLanguage || window.navigator.language;
+            }
+            // default
+            if (isValid(language)) {
+                return displayLang;
+            } else {
+                return DEFAULT_LANG;
+            }
+        };
+
+        /**
+         * Helper function to determine format constant.
+         *
+         * Defaults to DEFAULT_FORMAT if malformed or unpassed.
+         *
+         * @param {object} format (optional) A format constant.
+         * @returns {string} A format constant.
+         */
+        getDisplayFormat = function(format) {
+            var displayFormat = DEFAULT_FORMAT;
+            if (isValid(format)) {
+                displayFormat = format;
+            }
+            return displayFormat;
+        };
+
+        /**
+         * Helper function to determine variable/context validity.
+         *
+         * @param {string} candidateVariable to test
+         * @returns {boolean}
+         */
+        isValid = function(candidateVariable) {
+            return candidateVariable !== undefined
+                && candidateVariable !== ''
+                && candidateVariable !== 'Invalid date'
+                && candidateVariable !== 'None';
+        };
+
+        DateUtils = {
+            dateFormatEnum: dateFormatEnum,
+            localize: localize,
+            stringToMoment: stringToMoment,
+            getSysTimezone: getSysTimezone,
+            localizeTime: localizeTime
+        };
+        return DateUtils;
+    });
+}).call(
+    this,
+    // Pick a define function as follows:
+    // 1. Use the default 'define' function if it is available
+    // 2. If not, use 'RequireJS.define' if that is available
+    // 3. else use the GlobalLoader to install the class into the edx namespace
+    // eslint-disable-next-line no-nested-ternary
+    typeof define === 'function' && define.amd ? define :
+        (typeof RequireJS !== 'undefined' ? RequireJS.define :
+            edx.GlobalLoader.defineAs('DateUtils', 'edx-ui-toolkit/js/utils/date-utils'))
+);
+
--- a/src/js/utils/specs/date-utils-spec.js
+++ b/src/js/utils/specs/date-utils-spec.js
@@ -1,0 +1,147 @@
+define(
+    [
+        '../../utils/spec-helpers/spec-helpers.js',
+        '../date-utils.js',
+        'moment',
+        'moment-timezone'
+    ],
+    function(
+        SpecHelpers,
+        DateUtils,
+        moment,
+        momentTZ
+    ) {
+        'use strict';
+
+        describe('DateUtils', function() {
+            beforeEach(function() {
+                setFixtures('<div class="test"></div>');
+            });
+
+            describe('stringToMoment', function() {
+                it('returns an empty string if provided no time', function() {
+                    expect(DateUtils.stringToMoment()).toEqual(undefined);
+                });
+                it('converts a string to a moment.js object', function() {
+                    expect(
+                        DateUtils.stringToMoment('2016-10-14 08:00:00+00:00').format('ll')
+                    ).toEqual('Oct 14, 2016');
+                });
+            });
+
+            describe('getSysTimezone', function() {
+                it('attempts to determine system timezone', function() {
+                    // we need to determine the system tz offset from which to compare
+                    var sysDate = new Date();
+                    var sysDateOffset = sysDate.getTimezoneOffset();
+                    var dateUtilOffset = momentTZ.tz.zone(DateUtils.getSysTimezone()).offset(sysDate);
+                    expect(sysDateOffset).toEqual(dateUtilOffset);
+                });
+            });
+
+            describe('localizeTime', function() {
+                it('tests error conditions', function() {
+                    expect(DateUtils.localizeTime()).toEqual('');
+                });
+                it('tests sys time determiner', function() {
+                    var testDate = DateUtils.stringToMoment('2016-10-14 08:00:00+00:00');
+                    expect(
+                         moment(DateUtils.localizeTime(testDate).format()).isValid()
+                         ).toBeTruthy();
+                });
+                it('offsets time according to a determined timezone', function() {
+                    var testDate = DateUtils.stringToMoment('2016-10-14 08:00:00+00:00');
+                    // in order to avoid DST messes, we'll just pick some TZs without
+                    var TestZones = {
+                        'America/Winnipeg': 'Oct 14, 2016 3:00 AM',
+                        'Pacific/Honolulu': 'Oct 13, 2016 10:00 PM',
+                        'Asia/Ho_Chi_Minh': 'Oct 14, 2016 3:00 PM',
+                        'Atlantic/Reykjavik': 'Oct 14, 2016 8:00 AM',
+                        RANDOM_STRING: 'Oct 14, 2016 8:00 AM'
+                    };
+                    Object.keys(TestZones).forEach(function(key) {
+                        expect(
+                            new DateUtils.localizeTime(testDate, key).format('lll')
+                        ).toEqual(TestZones[key]);
+                    });
+                });
+            });
+
+            describe('localize', function() {
+                it('tests error conditions', function() {
+                    var context = {};
+                    expect(
+                        DateUtils.localize(context)
+                    ).toEqual('');
+                    context = {
+                        datetime: 'RANDOM_STRING',
+                        timezone: 'RANDOM_STRING',
+                        language: 'RANDOM_STRING'
+                    };
+                    expect(DateUtils.localize(context)).toEqual('');
+                });
+                it('tests defaults', function() {
+                    var context = {
+                        datetime: '2016-10-14 08:00:00+00:00',
+                        timezone: 'RANDOM_STRING',
+                        language: 'RANDOM_STRING'
+                    };
+                    expect(DateUtils.localize(context)).toEqual('Oct 14, 2016 08:00 UTC');
+                });
+
+                it('generates a localized string based on a language preference', function() {
+                    var TestLangs = {
+                        en: 'Oct 14, 2016 08:00 UTC',
+                        ru: '14 окт. 2016 г. 08:00 UTC',
+                        ar: '١٤ تشرين الأول أكتوبر ٢٠١٦ ٠٨:٠٠ UTC'
+                    };
+                    Object.keys(TestLangs).forEach(function(key) {
+                        var context = {
+                            datetime: '2016-10-14 08:00:00+00:00',
+                            timezone: 'UTC',
+                            language: key
+                        };
+                        expect(DateUtils.localize(context)).toEqual(TestLangs[key]);
+                    });
+                });
+                it('checks for formatting', function() {
+                    var TestFormats = {
+                        time: '8:00:00 AM UTC',
+                        shortDate: 'Oct 14, 2016',
+                        longDate: 'Friday, October 14, 2016 8:00 AM',
+                        undefined: 'Oct 14, 2016 08:00 UTC'
+                    };
+                    Object.keys(TestFormats).forEach(function(key) {
+                        var context = {
+                            datetime: '2016-10-14 08:00:00+00:00',
+                            timezone: 'UTC',
+                            language: 'en',
+                            format: DateUtils.dateFormatEnum[key]
+                        };
+                        expect(DateUtils.localize(context)).toEqual(TestFormats[key]);
+                    });
+                });
+
+                it('generates a localized datetime string', function() {
+                    var context = {
+                        datetime: '2016-10-14 08:00:00+00:00',
+                        timezone: 'Atlantic/Azores'
+                    };
+                    expect(DateUtils.localize(context)).toEqual('Oct 14, 2016 08:00 AZOST');
+                });
+
+                it('generates a more complex localized datetime string', function() {
+                    var context = {
+                        datetime: '2016-10-14 08:00:00+00:00',
+                        timezone: 'Pacific/Kiritimati',
+                        language: 'ar',
+                        format: DateUtils.dateFormatEnum.longDate
+                    };
+                    expect(
+                        DateUtils.localize(context)
+                    ).toEqual('الجمعة ١٤ تشرين الأول أكتوبر ٢٠١٦ ٢٢:٠٠');
+                });
+            });
+        });
+    }
+);
--- a/.gitignore
+++ b/.gitignore
@@ -60,3 +60,5 @@ node_modules
 _preview_site
 _site
 _tmp_preview_config.yml
+
+coverage/coverage-js/
--- a/AUTHORS
+++ b/AUTHORS
@@ -2,3 +2,4 @@ Dennis Jen <djen@edx.org>
 Alasdair Swan <aswan@edx.org>
 Andy Armstrong <andya@edx.org>
 Alisan Tang <atang@edx.org>
+Gregory Martin <greg@edx.org>
--- a/bower.json
+++ b/bower.json
@@ -22,6 +22,8 @@
     "sinon": "~1.17.3",
     "text": "~2.0.12",
     "underscore": "~1.8.3",
-    "urijs": "~1.16.1"
+    "urijs": "~1.16.1",
+    "moment": "^2.15.1",
+    "moment-timezone": "~0.5.5"
   }
 }
--- a/test/karma.conf.js
+++ b/test/karma.conf.js
@@ -54,6 +54,8 @@ module.exports = function(config, overrideOptions) {
             {pattern: 'node_modules/requirejs-text/text.js', included: false},
             {pattern: 'node_modules/sinon/**/*.js', included: false},
             {pattern: 'node_modules/urijs/src/*.js', included: false},
+            {pattern: 'node_modules/moment/min/moment-with-locales.min.js', included: false},
+            {pattern: 'node_modules/moment-timezone/builds/moment-timezone-with-data.js', included: false},
             {pattern: 'test/jquery.simulate.js', included: false},
 
             // register all the UI Toolkit source and Underscore templates
--- a/test/require-config.js
+++ b/test/require-config.js
@@ -13,6 +13,8 @@ require.config({
         text: 'node_modules/requirejs-text/text',
         underscore: 'node_modules/underscore/underscore',
         sinon: 'node_modules/sinon/lib/sinon',
+        moment: 'node_modules/moment/min/moment-with-locales.min',
+        'moment-timezone': 'node_modules/moment-timezone/builds/moment-timezone-with-data',
 
         // URI and its dependencies
         URI: 'node_modules/urijs/src/URI',


bjacobel:
@yro You probably don't have access to the S3 bucket for doc previews so I built one for you: http://ux-test.edx.org/yro_datetime-utils/utilities/date-utils/

I'll review this later today.

yro:
@bjacobel Thanks! I see some cleanup in the docs that I can do already.

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,292 @@
...
+         * local timezone formatted to the user's preferred language
+         *
+         * Most of the heavy lifting is done by the 'moments' js library,
+         * with the moment-with-locales and moment-timezones add-ons
+         *
+         * by sending preferences data, even in an incomplete state, we can generate the appropriate
+         * datetime for surfacing on the frontend
+         *
+         * for example, if you're a French speaker currently accessing the site in Rangoon,
+         * who you have not set your timezone preferences:
```
efischer19: Nit of the super small variety: this sentence is grammatically weird.

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,292 @@
...
+        UserTimeZone = function(timezone) {
+            // linter wants a different variable
+            var returnzone;
+            if (timezone === undefined || timezone === 'None') {
+                returnzone = MomentTZ.tz.guess();
+            } else {
+                returnzone = timezone;
+            }
+            if (returnzone === undefined) {
+                return 'UTC';
```
efischer19: You can define `returnzone` as `'UTC'` initially, then just let it be re-assigned (or not) as needed.

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,292 @@
...
+
+            if (language === undefined || language === 'None') {
+                displaylanguage = window.navigator.userLanguage || window.navigator.language;
+            } else {
+                displaylanguage = language;
+            }
+
+            /** default to en-us */
+            if (language === undefined) {
+                displaylanguage = 'en-US';
```
efischer19: Same type of comment here - `displaylanguage` is going through initialize as undefined->if some stuff assign a value->if still undefined use this default, you could go through initialize with default -> if some stuff assign a different value.

Also, consider using camelCase on your multi-word variable names for readability.

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,289 @@
...
+         * convert sent string to mutable 'moment' datetime object
+         *
+         * @param {string} datetime, can be just a date,
+         * or a complex date/time string with TZ offset. This is fairly open, and
+         * many strings will satisfy this (01-01-16 and Jan 1, 2016 will both work)
+         *
+         * @returns {object} a moment.js datetime object in UTC
+         *
+        */
+        DateObject = function(datetime) {
```
nasthagiri: Should this be renamed to `UTCTime` since that's what it's returning?

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,289 @@
...
+        UserTimeZone = function(timezone) {
+            // linter wants a different variable
+            var returnzone;
+            if (timezone === undefined || timezone === 'None') {
+                returnzone = MomentTZ.tz.guess();
+            } else {
+                returnzone = timezone;
+            }
+            if (returnzone === undefined) {
+                return 'UTC';
```
nasthagiri: But what if `MomentTZ.tz.guess` returns `undefined`?  We'll need to default to `UTC` in that case, right?

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,289 @@
...
+         * @param {object} localizedtime a moment.js datetime object in a determined timezone
+         * @param {string} language: an ISO 639-1 language string (e.g. 'en', 'fr')
+         * @param {string} format: (optional) a desired format constant
+         *
+         * @return {string} a language/timezone aware localized string
+        */
+        TimeDisplay = function(localizedtime, language, format) {
+            /**
+             * This is a function to determine the language based on browser settings,
+             * but should probably be overridden by platform-set language preference
```
nasthagiri: What do you mean by "platform-set language preference"?  In your POC, the moment library was able to pick up the local device settings as well.

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,289 @@
...
+            }
+            /**
+             * Determine the desired format,
+             * default to DEFAULT_DATE_TIME_FORMAT if undefined
+             *
+             */
+            if (typeof format !== undefined) {
+                displayformat = DATE_FORMATS[format];
+            }
+            if (typeof format === undefined || format === undefined || format === 'None') {
```
nasthagiri: why is this not an if-else statement?  And the second if statement has more conditions than the first.

yro: Yeah, that's not right -- I'll fix it.

src/js/utils/date-utils.js
yro: That's awkwardly worded. I meant that user pref-lang override the browser settings (like in the timezones). Fixed.

src/js/utils/date-utils.js
yro: I thought, since it's objectifying the datetime string, that this would work?

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,289 @@
...
+            localizedtime = LocalizedTime(
+                DateUtils.DateObject(context.datetime),
+                UserTimeZone(context.timezone)
+                );
+            displaytime = TimeDisplay(localizedtime, context.language, context.format);
+
+            if (context.modifier === undefined || context.modifier.length === 0) {
+                return displaytime;
+            }
+            /**
```
bjacobel: A normal block comment (just `/*` `*/`) should be fine here, we only use the JSDoc style for function definitions.

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,289 @@
...
+             * Optional handler for static strings via
+             * StringUtils in the ui toolkit
+             *
+             * Please Note: the preferred method is via underscore in the mako
+             * templates, however this will also work.
+            */
+            displaystring = StringUtils.interpolate(
+                context.modifier + ' {date}', {date: displaytime}
+                );
```
bjacobel: Same on the unindent here

bjacobel: Also - haven't used StringUtils.interpolate in a while, but I think this can just be `StringUtils.interpolate('{modifier} {date}', {date: displaytime, modifier: context.modifier})`

src/js/utils/specs/date-utils-spec.js
```diff
@@ -0,0 +1,163 @@
...
+                });
+                it('convert a string to an object', function() {
+                    expect(DateUtils.DateObject('2016-10-14 08:00:00+00:00').format('ll')).toEqual('Oct 14, 2016');
+                });
+            });
+
+            describe('UserTimeZone', function() {
+                it('attempt to determine users timezone', function() {
+                    /**
+                     * we need to determine the system tz offset from which to compare
```
bjacobel: Can just be a one-line comment

src/js/utils/specs/date-utils-spec.js
```diff
@@ -0,0 +1,163 @@
...
+                it('returns an empty string if provided no time', function() {
+                    expect(DateUtils.DateObject()).toEqual('');
+                });
+                it('convert a string to an object', function() {
+                    expect(DateUtils.DateObject('2016-10-14 08:00:00+00:00').format('ll')).toEqual('Oct 14, 2016');
+                });
+            });
+
+            describe('UserTimeZone', function() {
+                it('attempt to determine users timezone', function() {
```
bjacobel: `it('attempts`

src/js/utils/specs/date-utils-spec.js
```diff
@@ -0,0 +1,163 @@
...
+                    /**
+                     * we need to determine the system tz offset from which to compare
+                     *
+                     */
+                    var d = new Date();
```
bjacobel: Avoid single-letter variable names, give them a descriptive name if possible

src/js/utils/specs/date-utils-spec.js
```diff
@@ -0,0 +1,163 @@
...
+                it('will take timezone input', function() {
+                    expect(DateUtils.UserTimeZone('America/New_York')).toEqual('America/New_York');
+                });
+            });
+
+            describe('LocalizedTime', function() {
+                it('test error conditions', function() {
+                    var t = DateUtils.DateObject('2016-10-14 08:00:00+00:00');
+                    expect(
+                        DateUtils.LocalizedTime()).toEqual('');
```
bjacobel: If this needs to be broken across multiple lines, it should be three lines like so:

```javascript`
expect(
   foo
).toEqual('')
```

But it looks like this would fit under the line length limit if it was all just one line.

src/js/utils/specs/date-utils-spec.js
```diff
@@ -0,0 +1,163 @@
...
+                    expect(
+                        DateUtils.LocalizedTime(t).format('lll')
+                    ).toEqual('Oct 14, 2016 8:00 AM');
+                });
+                it('advance or retreat time according to a determined timezone', function() {
+                    var t = DateUtils.DateObject('2016-10-14 08:00:00+00:00');
+                    /**
+                     * in order to avoid DST messes, we'll just pick some TZs without
```
bjacobel: Single line comment

src/js/utils/specs/date-utils-spec.js
```diff
@@ -0,0 +1,163 @@
...
+                    var f = MomentTZ.tz.zone(DateUtils.UserTimeZone()).offset(d);
+                    expect(e).toEqual(f);
+                });
+                it('will take timezone input', function() {
+                    expect(DateUtils.UserTimeZone('America/New_York')).toEqual('America/New_York');
+                });
+            });
+
+            describe('LocalizedTime', function() {
+                it('test error conditions', function() {
```
bjacobel: Make sure that the strings in these `describe` / `it` unroll into a sentence, like "LocalizedTime tests error conditions" so that the error messages when tests fail make sense.

yro:
@bjacobel @nasthagiri @efischer19 I believe this is ready for a secondary review.

src/js/utils/date-utils.js
```diff
@@ -1,42 +1,42 @@
...
-
-/**
- * NOTE: There's a dependency with moment-timezone which relies on prepackaged timezone data,
- * and may not reliably install with npm. I've attempted to correct it, but need feedback as to
- * whether or not it's working on other's machines. It will fail tests if it's not working.
+ * Most of the heavy lifting is done by the 'moments' js library,
+ * with the moment-with-locales and moment-timezones add-ons.
  *
- *  Avail Here: http://momentjs.com/downloads/moment-timezone-with-data.js
+ * by sending preferences data, even in an incomplete state, we can generate the appropriate
```
efischer19: Microscopic nit: capitalize `By`

src/js/utils/date-utils.js
```diff
@@ -56,224 +56,215 @@ var DATE_FORMATS = {
...
-         * var context = {
-         *    datetime: '2016-10-14 08:00:00+00:00',
-         *    language: "fr",
-         *    }
-         * DateUtils.localize(context)
-         *
-         * which will render: '14 oct. 2016 14:30 MMT'
+         * Helper function to determine variable/context validity.
          *
+         * @param {string} variable to test
```
efischer19: `variable` could have a more descriptive name, like `possibleDate` or something.

src/js/utils/date-utils.js
```diff
@@ -56,224 +56,215 @@ var DATE_FORMATS = {
...
-         * @param {string} datetime, can be just a date,
-         * or a complex date/time string with TZ offset. This is fairly open, and
-         * many strings will satisfy this (01-01-16 and Jan 1, 2016 will both work)
+         * The input datetimeString can be just a date, a date and time,
+         * or a complex date/time string with TZ offset. This is fairly wide-open, as
+         * many strings will satisfy this requirement.
+         * (e.g. 01-01-16 and Jan 1, 2016, and 2016-01-01 01:00:00+00:00 will work)
          *
-         * @returns {object} a moment.js datetime object in UTC
+         * Recommend using something fully informant, like '2016-01-01 01:00:00+00:00',
```
efischer19: Is `informant` the right word here, or should it be `qualified`? I always though an informant was a person, but maybe that's my vocabulary lesson for today...

src/js/utils/date-utils.js
yro: Done. candidateVariable (since it's not always a date).

src/js/utils/date-utils.js
yro: That's fair -- "It is recommended to use a fully informed string, e.g. '2016-01-01 01:00:00+00:00', to limit possible date, time, and timezone ambiguities."

src/js/utils/date-utils.js
yro: Fixed.

package.json
```diff
@@ -22,6 +22,8 @@
     "backbone.paginator": "~2.0.3",
     "gulp-shell": "^0.5.2",
     "jquery": "~2.2.0",
+    "moment": "^2.15.1",
+    "moment-timezone": "~0.5.5",
```
bjacobel: This is a good Q - I think `^` for moment and `~` for moment-timezone is the right choice here because ^ will pick up all changes to the minor version (ie, everything up to 3.0.0) but ~ will limit changes to the patch version (ie, it will pick up 0.5.6, but it will not pick up 0.6) - because the API might have changed there.

src/js/utils/date-utils.js
yro: Or, in a less dictatorial wording "The recommendation is...". At any rate, fixed.

yro:
Tagging @cptvitamin for a quick a11y onceover.

cptvitamin:
LGTM 👍

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,283 @@
...
+         * Attempt to determine user's timezone from preferences, then browser settings.
+         * Will default to UTC.
+         *
+         * @param {string} timezone 'Region/Population_Center' e.g. 'America/New_York'.
+         * If this is None or undefined, we attempt to determine the TZ from system/browser settings.
+         * @return {string} Will attempt to return, in this order, the entered, determined, and default (UTC)
+         * timezone.
+         *
+         */
+        userTimezone = function(timezone) {
```
nasthagiri: What is a more accurate name for this function?  Maybe `localTimezoneOrUTC`?
There is no "user", per say, that is involved here.  Since it includes "user" in its name I assumed it checked user preferences.  But this general purpose function simply guesses the local timezone and if not found, returns UTC.

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,283 @@
...
+            var determinedZone;
+            if (isValid(timezone)) {
+                determinedZone = timezone;
+            } else {
```
nasthagiri: You should consider having the caller check the validity of the timezone rather than having this function do 2 things at once.  Since this is in a toolkit, it's better to have low-level utility functions.  So if I just want to get the local timezone, I should be able to do so without passing in a timezone.  Although the docstring for this function does say that the timezone can be null, it's unclear from the function signature that it's an optional parameter.

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,283 @@
...
+
+            if (isValid(determinedZone) === false) {
+                determinedZone = 'UTC';
+            }
+            return determinedZone;
+        };
+
+        /**
+         *
+         * Attempt to offset original time based on timezone.
```
nasthagiri: Consider changing this to: "Returns a new time that is offset by the requested timezone."

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,283 @@
...
+         *
+         * @param {Object} dateTimeObject A moment.js datetime object.
```
nasthagiri: Consider removing all "Object" prefixes from your function and variable names, as they don't add much value.  It is assumed that all objects that are passed around are objects.  Rather, if any are specifically, strings, you can include "string" in the name - and have the default assumption be "Object".

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,283 @@
...
+         * @param {string} userTZ A timezone representation, e.g. 'America/New_York'.
+         * @return {object} A moment.js datetime object in a determined timezone.
+         */
+        localizeTime = function(dateTimeObject, userTZ) {
```
nasthagiri: nice name.  :)

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,283 @@
...
+                localTime = '';
+            }
+            return localTime;
+        };
+
+        /**
+         *
+         * Generate a timezone-aware and language-aware string.
+         *
+         * This includes a hook set of built-ins for standard formatting.
```
nasthagiri: This sentence seems superfluous.  Also, not sure what you mean by "hook set" here.

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,283 @@
...
+         *
+         * Language is currently overridden by platform-set language preference. If no preference,
```
nasthagiri: Not sure why this says the language is "currently overridden"?  This makes me think that the requested language is never used and instead overridden by the one set in the browser.  However, that's not what the code seems to do.

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,283 @@
...
+            }
+            displayTime = localTime.locale(displayLang).format(displayFormat);
+            // Attempt to repair to UTC/US
+            if (isValid(displayTime) === false) {
+                displayTime = localTime.tz('UTC').locale('en').format(displayFormat);
+            }
+            if (isValid(displayTime) === false) {
+                displayTime = '';
+            }
+            return displayTime;
```
nasthagiri: Consider splitting up this function into 3 helpers.  So in the end, this function would just be:
```
var displayLang = _getDisplayLang(language)
var displayFormat = _getDisplayFormat(format)
return _getDisplayTime(displayLang, displayFormat)
```

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,283 @@
+/**
+ * Create a formatted, locale-aware string representation of a datetime.
```
nasthagiri: It would be helpful to add an example of how a client would call the function(s) in this module.

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,283 @@
...
+            } else if (candidateVariable === 'undefined') {
+                valid = false;
+            } else if (candidateVariable === '') {
+                valid = false;
+            } else if (candidateVariable === 'Invalid date') {
+                valid = false;
+            } else {
+                valid = true;
+            }
+            return valid;
```
andy-armstrong: In general, you don't need to have conditionals that return booleans, because the conditional expressions themselves are the result you want. In addition, you can benefit from all the ways in which JavaScript strings are falsey (http://stackoverflow.com/questions/19839952/all-falsey-values-in-javascript).

I believe you can change this whole thing to:

    return candidateVariable && candidateVariable !== 'Invalid date';

I'd also make 'Invalid date' into a constant.

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,283 @@
+/**
+ * Create a formatted, locale-aware string representation of a datetime.
+ *
+ * These functions are meant to act as a general localization function
+ * for server-provided UTC datetimes, creating UI-friendly strings in the user's
+ * local timezone formatted to the user's preferred language
```
yro: I don't disagree. Fixed.

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,283 @@
...
+ *
+ * Most of the heavy lifting is done by the 'moments' js library,
+ * with the moment-with-locales and moment-timezones add-ons.
+ *
+ * By sending preferences data, even in an incomplete state, we can generate the appropriate
+ * datetime string for surfacing to the user.
```
yro: Yep, that's fair. I was too tunnel-visioned as to the use case.

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,283 @@
...
+ *
+ * @module DateUtils
+ */
+
+
+/*
+DATE FORMATS:
+These were previously called from edx-platform/common/djangoapps/util/date_utils.py,
+but in an effort to move this to the front-end, they are redefined here and translated into
+local-aware moment.js formats.
```
yro: Fixed.

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,283 @@
...
+DEFAULT_DATE_TIME_FORMAT
+example: Jan 01, 2016 15:30 UTC
+python: "%b %d, %Y %H:%M"
+*/
+
+var DATE_FORMATS = {
+    DEFAULT_SHORT_DATE_FORMAT: 'll',
+    DEFAULT_LONG_DATE_FORMAT: 'LLLL',
+    DEFAULT_TIME_FORMAT: 'LTS z',
+    DEFAULT_DATE_TIME_FORMAT: 'll HH[:]mm z'
```
yro: Yep!

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,283 @@
...
+        ) {
+        var DateUtils, isValid, dateObject, userTimezone, localizeTime, timeDisplay, localize;
+
+        /**
+         * Helper function to determine variable/context validity.
+         *
+         * @param {string} candidateVariable to test
+         * @returns {boolean}
+         */
+        isValid = function(candidateVariable) {
```
yro: Okay - Fixed.

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,283 @@
...
+        'edx-ui-toolkit/js/utils/string-utils',
+        'moment',
+        'momenttimezone'
+    ], function(
+        $,
+        StringUtils,
+        moment,
+        momentTZ
+        ) {
+        var DateUtils, isValid, dateObject, userTimezone, localizeTime, timeDisplay, localize;
```
yro: No, that's just some sawdust left from construction. Fixed.

src/js/utils/date-utils.js
yro: @andy-armstrong @nasthagiri Fancy, I like that. I found a bug where I wasn't properly testing undefined variables as well. Thanks!

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,283 @@
...
+         * (e.g. 01-01-16 and Jan 1, 2016, and 2016-01-01 01:00:00+00:00 will work)
+         *
+         * The recommendation is to use a fully informed string, e.g. '2016-01-01 01:00:00+00:00',
+         * to limit possible date, time, and timezone ambiguities.
+         *
+         * @param {string} datetimeString. A string representation of a datetime.
+         * @returns {object} A moment.js UTC datetime object.
+         *
+        */
+        dateObject = function(datetimeString) {
```
yro: Right, @nasthagiri  that makes sense. I actually went with 'stringtoObject', as it's a moment-defined object, and not a proper js datetime, per se.

src/js/utils/date-utils.js
yro: I went with 'timezoneOrUTC', as it's not necessarily a local timezone, possible a preferred timezone.

src/js/utils/date-utils.js
yro: Okay. Fixed.

src/js/utils/date-utils.js
yro: Fixed.

src/js/utils/date-utils.js
yro: I changed it, but it will get changed again in the next step of review.

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,283 @@
...
+            }
+            if (isValid(displayTime) === false) {
+                displayTime = '';
+            }
+            return displayTime;
+        };
+
+        /**
+         *
+         * This will attempt render a localized date/time in a
```
yro: fixed.

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,283 @@
...
+         *~~~
+         *
+         * - datetime: {string} The (UTC) datetime server-side string.
+         * - modifier: {string} (optional) 'due', 'deadline:', etc..
+         * - timezone: {string} (optional) the user set time zone preference, or 'None' if unset.
+         * This should be in a format like 'Region/Locale', e.g. 'Atlantic/Azores'. Defaults to 'UTC'.
+         * - language: {string} (optional), the user set language, an ISO 639-1 language string, defaults to 'en'
+         * - format: {string} (optional), a format constant. Defaults to 'DEFAULT_DATE_TIME_FORMAT'.
+         */
+        localize = function(context) {
```
yro: @nasthagiri I was under the impression that, stylistically, over, like three arguments should be passed in as a single argument. Should this be named something else?

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,283 @@
...
+                );
+                displayTime = timeDisplay(localTime, context.language, context.format);
+            } else {
+                displayTime = '';
+            }
+
+            if (isValid(context.modifier)) {
+                displayString = StringUtils.interpolate(
+                    '{modifier} {date}',
+                    {date: displayTime, modifier: context.modifier}
```
yro: Fixed.

src/js/utils/date-utils.js
yro: Okay, wasn't sure. Will do.

src/js/utils/date-utils.js
yro: @nasthagiri @andy-armstrong -- I have a question about this in general -- it's not stylistically consistent with the rest of the repo and it seems to cause the linter to be unhappy. I was able to get around it by the following: /* eslint no-underscore-dangle: ["error", { "allow": ["_getDisplayLang", "_getDisplayFormat", "_isValid"] }] */; is that legit?

src/js/utils/date-utils.js
yro: @nasthagiri will follow up.

src/js/utils/date-utils.js
yro: yep, fixed.

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,283 @@
+/**
+ * Create a formatted, locale-aware string representation of a datetime.
+ *
+ * These functions are meant to act as a general localization function
+ * for server-provided UTC datetimes, creating UI-friendly strings in the user's
+ * local timezone formatted to the user's preferred language
+ *
+ * Most of the heavy lifting is done by the 'moments' js library,
```
yro: Fixed.

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,283 @@
...
+         * This function will default to UTC/en-US display in default format in the case of
+         * an unresolvable language or timezone
+         *
+         * preferred_language = 'en' yields Oct. 14, 2016 09:00 BST
+         *
+         * preferred_language = 'ru' yields 14 окт 2016 г. 09:00 BST
+         *
+         * @param {object} localTime A moment.js datetime object offset to a determined timezone.
+         * @param {string} language An ISO 639-1 language string (e.g. 'en', 'fr').
+         * @param {string} format (optional) A desired format constant.
```
yro: Fixed.

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,283 @@
...
+            // Determine the desired format, default to DEFAULT_DATE_TIME_FORMAT
+            if (isValid(format) && format in DATE_FORMATS) {
+                displayFormat = DATE_FORMATS[format];
+            }
+            displayTime = localTime.locale(displayLang).format(displayFormat);
+            // Attempt to repair to UTC/US
+            if (isValid(displayTime) === false) {
+                displayTime = localTime.tz('UTC').locale('en').format(displayFormat);
+            }
+            if (isValid(displayTime) === false) {
```
yro: Fixed.

src/js/utils/date-utils.js
yro: Fixed :)

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,283 @@
...
+                    userTimezone(context.timezone)
+                );
+                displayTime = timeDisplay(localTime, context.language, context.format);
+            } else {
+                displayTime = '';
+            }
+
+            if (isValid(context.modifier)) {
+                displayString = StringUtils.interpolate(
+                    '{modifier} {date}',
```
yro: @andy-armstrong will follow up.

yro:
@andy-armstrong @nasthagiri I think this is ready for you two to look at again.

src/js/utils/date-utils.js
nasthagiri: gotcha.  I forgot that the use of `context` was a js convention.  thanks.

src/js/utils/date-utils.js
```diff
@@ -1,48 +1,39 @@
+/*
+eslint no-underscore-dangle: ["error", {
```
andy-armstrong: I spoke to @bjacobel about this and he reminded me that we explicitly lint against using leading and trailing underscores! Sorry for the conflicting advice, but let's remove them again. 

In fact, they aren't necessary at all in this class, because we are explicitly choosing what to make public in the returned object. If there are things we don't want folks to see, we just shouldn't return them.

src/js/utils/date-utils.js
```diff
@@ -169,103 +185,88 @@ var DATE_FORMATS = {
...
+                && candidateVariable !== 'None';
         };
 
         DateUtils = {
-            isValid: isValid,
-            dateObject: dateObject,
-            userTimezone: userTimezone,
+            localize: localize,
+            stringtoObject: stringtoObject,
+            _getSysTimezone: _getSysTimezone,
```
andy-armstrong: The intention of this object is to return the public methods. I'd remove all the functions that are currently named starting with underscore.

src/js/utils/date-utils.js
```diff
@@ -52,39 +43,71 @@ var DATE_FORMATS = {
...
+         * handler for a modifier_string (e.g. 'due', 'deadline:', etc.) via
+         * StringUtils.
+         *
+         * Please Note: the preferred method is via underscore in the mako
+         * templates, however this will also work.
+         *
+         * @param context
+         * @returns {string} A formatted, timezone-offset, internationalized date and time.
+         *
+         * - datetime: {string} The (UTC) datetime server-side string.
```
andy-armstrong: I think these lines are describing the context, so should be above the ``@returns`` line.

andy-armstrong:
@yro this looks great, but I still want to change the modifier to be more friendly to localization. I also changed my recommendation for internal variable names (sorry!).

src/js/utils/date-utils.js
yro: Fair enough. Fixed

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,284 @@
...
+            return displayFormat;
+        };
+
+        /**
+         * Helper function to determine variable/context validity.
+         *
+         * @param {string} candidateVariable to test
+         * @returns {boolean}
+         */
+        _isValid = function(candidateVariable) {
```
andy-armstrong: How can this get either the string "None" or "Invalid date"? They seem like strange values to be handling.

At least the first two clauses can be collapsed together, because undefined is falsey:

    return candidateVariable && candidateVariable !== 'Invalid date' && candidateVariable !== 'Invalid date'

yro: The moment library will return 'Invalid date' if passed malformed variables. I've also managed to get 'None' from django/mako. Just trying to handle novices.

src/js/utils/date-utils.js
yro: Fixed.

src/js/utils/date-utils.js
yro: Yep, fair enough. Fixed.

src/js/utils/specs/date-utils-spec.js
```diff
@@ -0,0 +1,157 @@
...
+                        timezone: 'Pacific/Kiritimati',
+                        language: 'ar',
+                        format: 'DEFAULT_LONG_DATE_FORMAT'
+                    };
+                    expect(
+                        DateUtils.localize(context)
+                    ).toEqual('الجمعة ١٤ تشرين الأول أكتوبر ٢٠١٦ ٢٢:٠٠');
+                });
+
+                it('tests StringUtil on a modifier string', function() {
```
andy-armstrong: I think you can remove this test now that modifiers have gone.

yro: Thanks Andy! Couldn't have done it without you.

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,258 @@
...
+        };
+
+        var DEFAULT_LANG = 'en';
+        var DEFAULT_TIMEZONE = 'UTC';
+        /**
+         *
+         * Renders a localized date/time in a standard format.
+         *
+         * Please Note: the preferred method is via underscore in the mako
+         * templates, however this will also work.
```
nasthagiri: ?? Why is the preferred method in the mako template?  I thought this was the new recommended method?

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,258 @@
...
+         * - timezone: {string} (optional) the user set time zone preference, or 'None' if unset.
+         * This should be in a format like 'Region/Locale', e.g. 'Atlantic/Azores'. Defaults to 'UTC'.
+         * - language: {string} (optional), the user set language, an ISO 639-1 language string, defaults to 'en'
+         * - format: {string} (optional), a format constant. Defaults to 'DEFAULT_DATE_TIME_FORMAT'.
+         *
+         * @returns {string} A formatted, timezone-offset, internationalized date and time.
+         *
+         */
+        localize = function(context) {
+            // made a little longer than necessary for clarity
```
nasthagiri: Is this comment now outdated?

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,258 @@
...
+            var displayTime = '';
+
+            if (isValid(context.datetime)) {
+                localTime = localizeTime(
+                    DateUtils.stringToMoment(context.datetime),
+                    context.timezone
+                );
+                displayTime = getDisplayTime(localTime, context.language, context.format);
+            }
+            return displayTime;
```
nasthagiri: Nice!  What an elegant function to enjoy reading.  ;)

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,258 @@
...
+            var displayTime = '';
+            var displayLang = getDisplayLang(language);
+            var displayFormat = getDisplayFormat(format);
+
+            if (isValid(displayLang) && isValid(displayFormat)) {
+                if (moment(localTime).isValid()) {
+                    displayTime = localTime.locale(displayLang).format(displayFormat);
+                } else {
+                    displayTime = '';
+                }
```
nasthagiri: Is this else clause needed?  `displayTime` is already defaulted to the empty string above.  So can remove these 3 lines, right?

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,258 @@
...
+            }
+            return displayTime;
```
nasthagiri: Question (not a comment): When could this function fail?  For example, is failure expected if moment cannot find the locale?  If there are expected failure conditions, is there a safe default we can provide other than an empty string?

src/js/utils/specs/date-utils-spec.js
```diff
@@ -0,0 +1,145 @@
...
+        describe('DateUtils', function() {
+            beforeEach(function() {
+                setFixtures('<div class="test"></div>');
+            });
+
+            describe('stringToMoment', function() {
+                it('returns an empty string if provided no time', function() {
+                    expect(DateUtils.stringToMoment()).toEqual('');
+                });
+                it('will convert a string to an moment.js object', function() {
```
nasthagiri: Nittiest of the nit: for consistency with the rest of the tests, phrase this in the present tense: "will convert" -> "converts"

nasthagiri: "an moment" -> "a moment"

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,258 @@
...
+         * @param {string} datetimeString. A string representation of a datetime.
+         * @returns {object} A moment.js UTC datetime object.
+         *
+        */
+        stringToMoment = function(datetimeString) {
+            var utcDateObject;
+            if (isValid(datetimeString)) {
+                utcDateObject = moment(datetimeString).utc();
+            } else {
+                utcDateObject = '';
```
nasthagiri: Instead of an empty string, shouldn't this return `undefined`?  As the contract of the function is that it returns an `object` and not a `string`.  Also, the docstring needs to be updated to indicate this.  (As this is an external function and you have a test for it as well.)

src/js/utils/specs/date-utils-spec.js
```diff
@@ -0,0 +1,145 @@
...
+    ) {
+        'use strict';
+
+        describe('DateUtils', function() {
+            beforeEach(function() {
+                setFixtures('<div class="test"></div>');
+            });
+
+            describe('stringToMoment', function() {
+                it('returns an empty string if provided no time', function() {
```
nasthagiri: See above comment about returning `undefined` instead since the function is expected to return an object.

src/js/utils/specs/date-utils-spec.js
```diff
@@ -0,0 +1,145 @@
...
+                });
+                it('will convert a string to an moment.js object', function() {
+                    expect(
+                        DateUtils.stringToMoment('2016-10-14 08:00:00+00:00').format('ll')
+                    ).toEqual('Oct 14, 2016');
+                });
+            });
+
+            describe('getSysTimezone', function() {
+                it('attempts to determine users timezone', function() {
```
nasthagiri: Nit: not necessarily "users", but the "system".  This general purpose toolkit could be used even server-side (in node.js, for example).

src/js/utils/specs/date-utils-spec.js
```diff
@@ -0,0 +1,145 @@
...
+                    // we need to determine the system tz offset from which to compare
+                    var sysDate = new Date();
+                    var sysDateOffset = sysDate.getTimezoneOffset();
+                    var dateUtilOffset = momentTZ.tz.zone(DateUtils.getSysTimezone()).offset(sysDate);
+                    expect(sysDateOffset).toEqual(dateUtilOffset);
+                });
```
nasthagiri: A contract of this public method is that it "falls back to UTC".  Did you plan to add a test to make sure that it does?

src/js/utils/specs/date-utils-spec.js
```diff
@@ -0,0 +1,145 @@
...
+                it('offsets time according to a determined timezone', function() {
+                    var testDate = DateUtils.stringToMoment('2016-10-14 08:00:00+00:00');
+                    // in order to avoid DST messes, we'll just pick some TZs without
+                    var TestZones = {
+                        'America/Winnipeg': 'Oct 14, 2016 3:00 AM',
+                        'Pacific/Honolulu': 'Oct 13, 2016 10:00 PM',
+                        'Asia/Ho_Chi_Minh': 'Oct 14, 2016 3:00 PM',
+                        'Atlantic/Reykjavik': 'Oct 14, 2016 8:00 AM'
+                    };
```
nasthagiri: great test cases.

src/js/utils/specs/date-utils-spec.js
```diff
@@ -0,0 +1,145 @@
...
+                        timezone: 'RANDOM_STRING',
+                        language: 'RANDOM_STRING'
+                    };
+                    expect(DateUtils.localize(context)).toEqual('Oct 14, 2016 08:00 UTC');
+                    context = {
+                        datetime: 'RANDOM_STRING',
+                        timezone: 'RANDOM_STRING',
+                        language: 'RANDOM_STRING'
+                    };
+                    expect(DateUtils.localize(context)).toEqual('');
```
nasthagiri: Nittiest of the nit: this should be moved to "tests error conditions" instead of "tests defaults", right?

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,258 @@
...
+
+        var DATE_FORMATS = {
+            DEFAULT_SHORT_DATE_FORMAT: 'll', // example: Jan 01, 2016
+            DEFAULT_LONG_DATE_FORMAT: 'LLLL', // example: Friday, January 01, 2016
+            DEFAULT_TIME_FORMAT: 'LTS z', // example: Jan 01, 2016 15:30 UTC
+            DEFAULT_DATE_TIME_FORMAT: 'll HH[:]mm z' // example: Jan 01, 2016 15:30 UTC
+        };
+
+        var DEFAULT_LANG = 'en';
+        var DEFAULT_TIMEZONE = 'UTC';
```
nasthagiri: Nit: Include the default format up here as well.  This way, all defaults will be declared together, including the default from the DATE_FORMATS.

src/js/utils/specs/date-utils-spec.js
```diff
@@ -0,0 +1,145 @@
...
+                        };
+                        expect(DateUtils.localize(context)).toEqual(TestLangs[key]);
+                    });
+                });
+                it('checks for formatting', function() {
+                    var TestFormats = {
+                        DEFAULT_TIME_FORMAT: '8:00:00 AM UTC',
+                        DEFAULT_SHORT_DATE_FORMAT: 'Oct 14, 2016',
+                        DEFAULT_LONG_DATE_FORMAT: 'Friday, October 14, 2016 8:00 AM'
+                    };
```
nasthagiri: Hmm.. why does the `DEFAULT_LONG_DATE_FORMAT` include the time?  The [example in your code](https://github.com/edx/edx-ui-toolkit/pull/85/files#diff-061b8b80c6ba6b948b73a2611b2897e7R51) doesn't.

src/js/utils/specs/date-utils-spec.js
```diff
@@ -0,0 +1,145 @@
...
+                        timezone: 'Pacific/Kiritimati',
+                        language: 'ar',
+                        format: 'DEFAULT_LONG_DATE_FORMAT'
+                    };
+                    expect(
+                        DateUtils.localize(context)
+                    ).toEqual('الجمعة ١٤ تشرين الأول أكتوبر ٢٠١٦ ٢٢:٠٠');
+                });
+            });
+        });
```
nasthagiri: Excellent test cases!

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,258 @@
...
+ *
+ *
+ * Example:
+ *
+ *~~~ javascript
+ * var context = {
+ *    datetime: '2016-10-14 08:00:00+00:00',
+ *    timezone: 'Pacific/Kiritimati',
+ *    language: 'ar',
+ *    format: 'DEFAULT_LONG_DATE_FORMAT'
```
yro: This is supposed to decode (is that the right word?) via the getDisplayFormat() method. Am I missing something?

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,258 @@
...
+ *    };
```
yro: Haha. Fair enough.

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,258 @@
...
+ * return_string = DateUtils.localize(context);
```
yro: sounds good.

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,258 @@
...
+ *~~~
```
yro: Yep!

src/js/utils/date-utils.js
yro: Yes, that's smart. Done.

src/js/utils/date-utils.js
yro: Oh, sorry that's an outdated comment related to StringUtils. Good catch. Gone.

src/js/utils/date-utils.js
yro: It is indeed.

src/js/utils/date-utils.js
yro: Ha! Thanks. I see it now.

src/js/utils/date-utils.js
yro: Yep!

src/js/utils/date-utils.js
yro: Yes, that makes sense. I was just going for 'empty', but undefined is better.

src/js/utils/date-utils.js
yro: I *think* I caught all the possible failure conditions. If moment fails to determine the locale or the timezone, we default to en/UTC, which is the current display condition, and if the formatting fails, we default to a safe format string. One possibility is passing through the original datetime argument? I'll follow up.

src/js/utils/specs/date-utils-spec.js
yro: yep!

src/js/utils/specs/date-utils-spec.js
yro: Gah! Good catch. Thanks.

src/js/utils/specs/date-utils-spec.js
yro: Yep!

src/js/utils/specs/date-utils-spec.js
yro: Yes, good call. Done.

src/js/utils/specs/date-utils-spec.js
yro: Ha! Thanks.

src/js/utils/specs/date-utils-spec.js
yro: Yep!

src/js/utils/specs/date-utils-spec.js
yro: Thanks :)

src/js/utils/specs/date-utils-spec.js
yro: Oh, typo. Got it.

src/js/utils/date-utils.js
nasthagiri: ah.  I didn't realize this was passed as a string, not an constant enum value.

nasthagiri: Per our offline conversation, consider changing the interface so callers can pass in an enum constant instead of a string - this allows issues in the caller to be caught by the interpreter - and not silently fail.
This change can be made as part of this PR or a separate PR.

yro:
@nasthagiri Ready for you to check again!

src/js/utils/date-utils.js
```diff
@@ -0,0 +1,255 @@
...
+            localize,
+            stringToMoment,
+            localizeTime,
+            getSysTimezone,
+            getDisplayTime,
+            getDisplayLang,
+            getDisplayFormat,
+            isValid;
+
+        var dateFormatEnum = Object.freeze({
```
nasthagiri: I learn something new everyday.

nasthagiri:
👍 Great work!  Remember to squash before merging.

PR #89, opened 2016-12-29
addtl' formats

yro:
## Description
[TNL-6107](https://openedx.atlassian.net/browse/TNL-6107)
(subtask)

Addition of two format constants needed for edx-platform and edx-ora2, as well as a correction to an example for a format. 

## Manual Test Plan

TODO: If your changes require manual tests, include instructions for any required manual tests, and any manual testing that has already been performed.

## Testing Checklist
- [ ] Write unit tests for all new features.
- [ ] Write accessibility (a11y) tests for all new UI.
- [ ] Manually test responsive behavior.

## Non-testing Checklist
- [ ] Consider any documentation your change might need, and which users will be affected by this change.
- [ ] Consult with the Pattern Library team if adding a new component.

## Post-review

## Reviewers
TODO: In this section, tag specific reviewers you know will want to have a look at this PR. You can use the checklist below for organization:

- [x] @bjacobel 
- [ ] @efischer19 

TODO: Consider not just code reviewers but also including reviewers from Docs, UX, Accessibility and Product, if applicable.

### DIFF
--- a/src/js/utils/date-utils.js
+++ b/src/js/utils/date-utils.js
@@ -49,8 +49,10 @@
         var dateFormatEnum = Object.freeze({
             shortDate: 'll', // example: Jan 01, 2016
             longDate: 'LLLL', // example: Friday, January 01, 2016
-            time: 'LTS z', // example: Jan 01, 2016 15:30 UTC
-            defaultFormat: 'll HH[:]mm z' // example: Jan 01, 2016 15:30 UTC
+            time: 'LTS z', // example: 15:30:32 UTC
+            defaultFormat: 'll HH[:]mm z', // example: Jan 01, 2016 15:30 UTC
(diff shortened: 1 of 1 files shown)


src/js/utils/date-utils.js
```diff
@@ -49,8 +49,10 @@
         var dateFormatEnum = Object.freeze({
             shortDate: 'll', // example: Jan 01, 2016
             longDate: 'LLLL', // example: Friday, January 01, 2016
-            time: 'LTS z', // example: Jan 01, 2016 15:30 UTC
-            defaultFormat: 'll HH[:]mm z' // example: Jan 01, 2016 15:30 UTC
+            time: 'LTS z', // example: 15:30 UTC
```
efischer19: Does `LTS` include seconds information as well? A quick glance at moment docs says it does, but it's not in the inline comment at all.

yro: D'oh! You're right. Will fix.

PR #92, opened 2017-04-06
Add alternative to deprecated keyCodes

bjacobel:
The keyCode event property has been deprecated. See [the MDN link](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) for
a description of how the new standard for key identifiers should be used.

I ran into this while working on https://github.com/edx/edx-platform/pull/14751 and moving some jQuery event triggers to native `KeyboardEvents`.

### DIFF
--- a/src/js/utils/constants.js
+++ b/src/js/utils/constants.js
@@ -7,20 +7,33 @@
         /**
          * Reusable constants.
          *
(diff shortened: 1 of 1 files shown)


bjacobel:
I'll hold off on merging this until Ari fixes the master failures and I can rebase, it's not urgent.

PR #95, opened 2017-04-20
eslint config update & lintfixes

arizzitano:
## Description
- Upgrade `eslint`
- Upgrade `eslint-config-edx` and `eslint-config-edx-es5`
- Fix resulting eslint errors

## Post-review
- [x] Squash


package.json
```diff
@@ -35,8 +35,9 @@
     "css-loader": "~0.23.1",
     "del": "~2.2.0",
     "edx-pattern-library": "~0.12.5",
-    "eslint": "~2.13.1",
-    "eslint-config-edx": "~1.2.1",
+    "eslint": "^3.16.0",
```
bjacobel: AFAIK we don't need an explicit ESLint dependency anymore now that we're on NPM 3 (although maybe the Yarn thing you found has implications for this?)

src/js/utils/spec-helpers/ajax-helpers.js
```diff
@@ -65,7 +65,7 @@ define(['sinon', 'underscore', 'URI'], function(sinon, _, URI) {
 
         requests.currentIndex = 0;
         requests.restore = function() {
-            if (xhr && xhr.hasOwnProperty('restore')) {
+            if (xhr && Object.prototype.hasOwnProperty.call(xhr, 'restore')) {
```
bjacobel: this is a little yucky - does `&& 'restore' in xhr` work here?

src/js/utils/specs/ajax-helpers-spec.js
```diff
@@ -34,7 +34,7 @@ define(
                     $.ajax(testUrl);
 
                     expect(requests instanceof Array).toBeTruthy();
-                    expect(requests.hasOwnProperty('currentIndex')).toBeTruthy();
+                    expect(Object.prototype.hasOwnProperty.call(requests, 'currentIndex')).toBeTruthy();
```
bjacobel: I'd like to avoid these Object.prototype.x.call calls if we can, see above

package.json
arizzitano: Yep, we do need it since Yarn won't link binaries for transitive deps and recommends either an explicit dependency or a postinstall creating the link manually. The explicit dependency seemed like a cleaner option in this case.

src/js/utils/specs/ajax-helpers-spec.js
arizzitano: Agreed this is gross. `'prop` in object` works here afaik and is much prettier :+1:

PR #401, opened 2024-05-24
feanil/remove ux pattern library

feanil:
- **feat: Remove the unused edx-pattern-library dev dependency.**
- **build: Bump the version manually.**

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "edx-ui-toolkit",
-  "version": "1.5.4",
+  "version": "1.5.5",
   "description": "A JavaScript toolkit for building edX user interfaces",
   "license": "Apache-2.0",
   "repository": {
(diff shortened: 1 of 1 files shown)


kdmccormick (approved):

kdmccormick (approved):

arbrandes (approved):
+1

kdmccormick (approved):

kdmccormick (approved):

kdmccormick:
@feanil some context, if it helps with the build failure at all: back in ~2016, when paragon superseded edx-pattern-library, it looks like edx-platform fully replaced its edx-pattern-library usage with [its own local shims](https://github.com/openedx/edx-platform/tree/master/common/static/sass/edx-pattern-library-shims). So with insights gone, and since neither edx-platform nor credentials use edx-pattern-library, I think we are safe to slash and burn any edx-pattern-library bits that we see.

PR #467, opened 2024-12-16
docs: Add a catalog-info.yaml file.

feanil:
Kyle will maintain this until we can get rid of it as a part of the
legacy UI deprecation.


feanil:
https://github.com/openedx/edx-ui-toolkit/pull/468 should resolve the test failures we're seeing here so this can be merged in after that has landed.

